PR #8, opened 2016-10-04
CBBD-141: Increase Wildfly connect timeouts

karlmdavis:
This seems to have been the root cause of [CBBD-141: Intermittent benchmark failure due to JBoss/Wildfly connect timeouts](http://issues.hhsdevcloud.us/browse/CBBD-141), and isn't a bad idea, regardless.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 
 # Ignore Maven output folders
 /target*/
+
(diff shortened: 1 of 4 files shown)


karlmdavis:
@dho937 and/or @sbrockhoff: Can I please get a review of this (tiny) change?

PR #9, opened 2016-10-11
Fixed bug in server-start.sh, where it called wrong -stop.sh on error.

karlmdavis:
Somewhat related to CBBD-138, as it was annoying me while I worked on it, but mostly just an unrelated bug that needed to be stomped. Should prevent us from having to restart build server every time an IT fails.


karlmdavis:
@dho937 Review?

dho937:
:+1:

PR #10, opened 2016-10-18
upgrade to dstu3

dho937:


karlmdavis:
@dho937 You should do the honors: you've put in way more time on this one. I'm just glad I could help.

PR #14, opened 2016-10-25
CBBF-40: Clean up bbonfhir-server project (round 2)

karlmdavis:
http://issues.hhsdevcloud.us/browse/CBBF-40

In this round, I've renamed the project, renamed the packages, renamed some of the classes, and bumped the version to `1.0.0-SNAPSHOT`. See the commit messages for details.


karlmdavis:
@dho937: Can you please review and approve?

I believe the build is only broken because I updated the project's name in the POM's `<scm/>` section. Once this PR is merged, I'll be renaming the project in GitHub, which should resolve that problem.

PR #18, opened 2016-11-08
changes to run local server on windows/cygwin

dho937:


dho937:
@karlmdavis @abdulaleem  Please review when you get a chance.  Thanks.

PR #25, opened 2017-05-10
CBBD-248: Make the number of server DB connections configurable

karlmdavis:
The previous/default approach of calculating a value based on the number
of processors is okay, but doesn't really allow for multi-system
environments where the FHIR server has less/more CPUs than the ETL
server.


karlmdavis:
Thanks for the review, gentlemen!

PR #31, opened 2017-06-05
CBBF-60: Dialing down the log spamming from ETL

karlmdavis:
Every create operation now runs a search (thanks to our `If-None-Exists` usage), and every one of those gets logged.

It doesn't appear to be slowing down the FHIR server any, but it will eventually eat a ton of disk space. And it makes the logs almost useless -- can't find anything else in there with all of the noise.

http://issues.hhsdevcloud.us/browse/CBBF-60


karlmdavis:
Bother!

This change seems to change the timing just enough to trigger a deadlock bug in HAPI's XML parsing code. Can't merge this until that's resolved: https://github.com/jamesagnew/hapi-fhir/issues/664

Until then, we'll have to deal with the log noise. (About 1.9M beneficiaries into the load, the log file is at 4GB.)

karlmdavis:
The upstream issue wasn't what I thought it was (see the discussion in it), and isn't blocking this merge. So: merging.

PR #32, opened 2017-06-15
CBBD-270: Disable saving of searches into the HAPI DB

karlmdavis:
These saved searches were landing in the `hfj_search` table.

Just adds extra load that we don't need right now (during the initial load).

http://issues.hhsdevcloud.us/browse/CBBD-270


karlmdavis:
Not needed anymore, given the new Dynamic ETL architecture. (But mostly because paging is completely disabled now.)

PR #36, opened 2017-08-09
CBBF-63 Customize and verify the server's CapabilityStatement

karlmdavis:
http://issues.hhsdevcloud.us/browse/CBBF-63


karlmdavis:
@ekivemark: Can you take a look at the test case here and see if it captures all of the pieces in the backend server's `CapabilityStatement` that you rely on in the frontend? If not, what's missing? (Other than `Coverage` support, of course.)

PR #38, opened 2017-08-28
reference coverage resource within eob

dho937:


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PartDEventTransformer.java
```diff
@@ -249,19 +248,6 @@ private static ExplanationOfBenefit transformClaim(PartDEvent claimGroup) {
 					TransformerConstants.CODING_SYSTEM_CCW_PHRMCY_SRVC_TYPE_CD, claimGroup.getPharmacyTypeCode());
 		}
 
-		Coverage coverage = new Coverage();
-		coverage.addIdentifier().setSystem(TransformerConstants.CODING_SYSTEM_PDE_PLAN_CONTRACT_ID)
-				.setValue(claimGroup.getPlanContractId());
```
karlmdavis: Looks like the change leaves `claimGroup.getPlanContractID()` and `claimGroup.getPlanBenefitPackageId()` unmapped. I think I'd stick them in extensions?

PR #46, opened 2017-11-17
CBBF-86: changed the access level modifiers for some of the Transform…

glovercloud:
…erUtils so the jmeter performance tests could reuse existing code.  Also updated the pom so a jar could be built that would allow the performance tests to easily use the TransformerUtils class.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -134,12 +134,12 @@
 	<build>
 		<pluginManagement>
 			<plugins>
-        <plugin>
-          <artifactId>maven-war-plugin</artifactId>
-          <configuration>
-            <attachClasses>true</attachClasses>
-          </configuration>
-        </plugin>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -134,6 +134,12 @@
 	<build>
 		<pluginManagement>
 			<plugins>
+        <plugin>
+          <artifactId>maven-war-plugin</artifactId>
+          <configuration>
+            <attachClasses>true</attachClasses>
+          </configuration>
+        </plugin>
```
karlmdavis: Can you match the formatting used elsewhere in the file here? (Tabs, so... apologies, I guess. lol)

glovercloud:
Karl, could you check if the formatting is acceptable now so we can merge this one?

PR #47, opened 2017-12-19
CBBF-99: Fix mapping to handle null procedure dates

karlmdavis:
This was blowing up on the synthetic data. Would have been blowing up in
production, too, if we'd tried to pull one of these claim types.

https://issues.hhsdevcloud.us/browse/CBBF-99


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CCWProcedure.java
```diff
@@ -57,7 +58,7 @@ public LocalDate getProcedureDate() {
 	 *         <code>icdCode</code> was present
 	 */
 	static Optional<CCWProcedure> from(Optional<String> icdCode, Optional<Character> icdVersionCode,
-			LocalDate procedureDate) {
+			Optional<LocalDate> procedureDate) {
 		if (!icdCode.isPresent())
 			return Optional.empty();
```
dho937: @karlmdavis   Curious why the code is blowing up with the above code in place..I must be missing something.

karlmdavis: Good question! It's because the `LocalDate` isn't an `Optional`, so in order to call this method, you've already called its `.get()` method -- causing it to blow up if the date isn't present.

PR #55, opened 2018-01-17
Cbbf 114 abstract common fields in inpatient/snf

glovercloud:


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -516,4 +528,287 @@ static void validatePeriodDates(Optional<LocalDate> dateFrom, Optional<LocalDate
...
+	 *            an {@link Optional}&lt;{@link BigDecimal}&gt; shared field
+	 *            representing the claim PPS capital IME amount for the claim
+	 * @param claimPPSCapitalOutlierAmount
+	 *            an {@link Optional}&lt;{@link BigDecimal}&gt; shared field
+	 *            representing the claim PPS capital outlier amount for the claim
+	 * @param claimPPSOldCapitalHoldHarmlessAmount
+	 *            an {@link Optional}&lt;{@link BigDecimal}&gt; shared field
+	 *            representing the claim PPS old capital hold harmless amount for
+	 *            the claim
+	 */
```
dho937: This is very minor but Karl wanted me to also include the field names from CCW (names with underscores in them from InpatientClaimColumn) to be beside the above parameter names.  I know..a very tedious process but just wanted to share what I was doing.   Do this for any common methods you have in TransformerUtils and TransformerTestUtils.   Love the field descriptions.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -516,4 +528,287 @@ static void validatePeriodDates(Optional<LocalDate> dateFrom, Optional<LocalDate
...
+
+		// deductibleAmount
+		if (deductibleAmount != null) {
+			bc = new BenefitComponent(
+					TransformerUtils.createCodeableConcept(TransformerConstants.CODING_BBAPI_BENEFIT_BALANCE_TYPE,
+							TransformerConstants.CODED_BENEFIT_BALANCE_TYPE_DEDUCTIBLE));
+			bc.setAllowed(new Money().setSystem(TransformerConstants.CODED_MONEY_USD).setValue(deductibleAmount));
+			benefitBalances.getFinancial().add(bc);
+		}
+
```
dho937: Can you add a general FIXME comment to this method that basically says to review any of these numeric fields that we are checking for null to see if they indeed can be nullable.   A lot of our amount fields are non-nullable so we don't need the "if" check.   If the amount is 0; then we should show 0 to the end user.  Thanks.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/InpatientClaimTransformer.java
```diff
@@ -425,13 +285,12 @@ private static ExplanationOfBenefit transformClaim(InpatientClaim claimGroup) {
 					.setCode(TransformerConstants.CODED_MONEY_USD).setValue(claimLine.getTotalChargeAmount());
 
 			item.addAdjudication()
-					.setCategory(
-							TransformerUtils.createCodeableConcept(TransformerConstants.CODING_CCW_ADJUDICATION_CATEGORY,
-									TransformerConstants.CODED_ADJUDICATION_NONCOVERED_CHARGE))
+					.setCategory(TransformerUtils.createCodeableConcept(
+							TransformerConstants.CODING_CCW_ADJUDICATION_CATEGORY,
+							TransformerConstants.CODED_ADJUDICATION_NONCOVERED_CHARGE))
```
karlmdavis: Looks like spurious whitespace changes. No big deal, just try to avoid them in the future if you can.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -534,6 +539,305 @@ static void validatePeriodDates(Optional<LocalDate> dateFrom, Optional<LocalDate
...
+	 *            the claim
+	 * @param admittingLabel
+	 *            a {@link DiagnosisLabel} shared field representing the admitting
+	 *            label for the claim
+	 */
+	static void extractCommonDiagnosesInpatientSNF(Consumer<Optional<Diagnosis>> diagnosisAdder,
+			Optional<String> diagnosisAdmittingCode, Optional<Character> diagnosisAdmittingCodeVersion,
+			DiagnosisLabel admittingLabel) {
+		diagnosisAdder.accept(Diagnosis.from(diagnosisAdmittingCode, diagnosisAdmittingCodeVersion, admittingLabel));
+	}
```
karlmdavis: I'm not sure that the above method is adding any value. Do you think it would simplify the code to remove it?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/InpatientClaimTransformer.java
glovercloud: That's the Eclipse formatter making it correct, I believe the old spurious spaces was actually incorrect.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
glovercloud: Yeah, I just added it because it was flagged in Dave's spreadsheet.  Its real value is probably if more common diagnosis between these claim types is found, otherwise not so much.

glovercloud:
Karl, the test fixes you are referring to where fixed and merge to master [here](https://github.com/CMSgov/bluebutton-data-server/pull/60).  I remerged master to this branch and the tests are working again so it should be good to go in this regard.

glovercloud:
Dave, can you check this before I merge back to master?  I merged the master to this branch with your latest changes but it was most difficult and I want to make sure I didn't inadvertently remove something you recently added.  I think it is good, the tests run, but I need another set of eyes on it.

PR #56, opened 2018-01-18
CBBF-119: Consolidated the set provider number calls from HHA, Hospic…

glovercloud:
…e, Inpatient, Outpatient and SNF claim types into a single method located in TransforerUtils.  Also updated corresponding tests.  Since DME does not have a provider number at the group level, a TODO comment was added in the relevant code and test sections to research how this is to be handled for that specific claim type.


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/DMEClaimTransformer.java
```diff
@@ -65,7 +65,15 @@ private static ExplanationOfBenefit transformClaim(DMEClaim claimGroup) {
...
-
+		
+		/*
+		 * TODO: DME does not have a provider number at the EOB level to map to but has
+		 * provider billing numbers at the claim line level. Need to do some research on
+		 * how this should be mapped, if it even can be, like the other claim types:
+		 * 
+		 * TransformerUtils.setProviderNumber(eob, claimGroup.getProviderNumber());
+		 */
+
```
karlmdavis: If we're not mapping that right now, that's a pretty big gap. Can you file a JIRA ticket for it and drop it into the next sprint?

glovercloud: [CBBF-126](https://issues.hhsdevcloud.us/browse/CBBF-126)

PR #57, opened 2018-01-19
CBBI-289: Updated the Jenkinsfile for new Jenkins server.

karlmdavis:
Some functional changes here:

* For now, run `mvn install` instead of `mvn deploy`. Will switch if/when a Nexus server is available.
* Only change versions for non-master builds.
* Archive the things that matter, but not binaries, as that's a duplicate waste of space. Do fingerprint binaries, though.

https://issues.hhsdevcloud.us/browse/CBBI-289

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,60 +1,66 @@
 /**
  * <p>
  * This is the script that will be run by Jenkins to build and test this 
- * project. This drives the project's continuous integration, delivery, and
- * deployment.
+ * project. This drives the project's continuous integration and delivery.
  * </p>
  * <p>
- * This script is run by Jenkins' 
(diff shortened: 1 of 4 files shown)


karlmdavis:
@dho937 and @glovercloud Added another commit here, and it's kind of a big one. I didn't want to manually check out and build that HAPI branch we were using on the Jenkins box, so I moved us to the first official release that contains it. 

All the tests still pass, but I'm also going to go through and manually verify that our API's responses haven't changed due to the upgrade.

Here's the change log for HAPI releases: <http://hapifhir.io/changes-report.html>.

Are you comfortable with this change?

karlmdavis:
Confirmed: with the `SAMPLE_A` resources loaded, there are no API response differences between this branch and `master` (for `Patient` or `ExplanationOfBenefit` -- didn't check `Coverage` because laziness).

PR #61, opened 2018-01-23
CBBF-100: script fixes that allow for successful builds of the data s…

glovercloud:
…erver on mac OSX.  This was essential to get working to support a deployment from a mac.

Karl, please verify the pkill change I made in the server stop script will not break anything on your side.  It was the only syntax I could get to work in OSX but wasn't sure if it was compatible with your platform.

### DIFF
--- a/bluebutton-server-app/src/main/config/server-stop.sh
+++ b/bluebutton-server-app/src/main/config/server-stop.sh
@@ -76,6 +76,6 @@ if [[ -z "${serverPids}" ]]; then
 	exit 0
 else
 	>&2 echo "Server processes still found. Sending KILL signal to all 'bluebutton-server' processes."
-	pkill KILL -f ".*java.*-Dbluebutton-server.*jboss-modules\.jar.*"
+	pkill -KILL -f ".*java.*-Dbluebutton-server.*jboss-modules\.jar.*"
 	>&2 echo "Server processes sent KILL signal."
 fi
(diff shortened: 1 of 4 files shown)


bluebutton-server-app/src/main/config/server-stop.sh
```diff
@@ -71,12 +70,12 @@ done
...
 # needed.
-serverPids=$(pgrep --full ".*java.*-Dbluebutton-server.*jboss-modules\.jar.*")
+serverPids=$(pgrep -f ".*java.*-Dbluebutton-server.*jboss-modules\.jar.*")
 if [[ -z "${serverPids}" ]]; then
 	echo "Server did actually stop."
 	exit 0
 else
 	>&2 echo "Server processes still found. Sending KILL signal to all 'bluebutton-server' processes."
-	pkill -KILL --full ".*java.*-Dbluebutton-server.*jboss-modules\.jar.*"
+	pkill KILL -f ".*java.*-Dbluebutton-server.*jboss-modules\.jar.*"
```
karlmdavis: Doesn't seem to be valid, per the man page on Ubuntu 14.04: <http://manpages.ubuntu.com/manpages/trusty/en/man1/pkill.1.html>. The signal name needs to have a leading dash, or else it's interpreted as the process pattern to match against.

And when running the build/ITs, I get the following output from `pkill`:

```
pkill: only one pattern can be provided
Try `pkill --help' for more information.
```

For that matter, it doesn't look valid for the Mac version of `pkill`, either: <https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/pkill.1.html>.

What happens if you have it as `pkill -KILL -f ...`?

glovercloud: I would swear the -KILL didn't work for me, but I retried it and it does.  Actually both versions seem to work on OSX but this branch should now have the -KILL version in it.

PR #59, opened 2018-01-22
CBBF-115-118-121: Diagnosis/Procedure/common field abstractions

dho937:

PR #62, opened 2018-01-24
Fixed line endings problem caused by earlier PR

karlmdavis:
Switching these files back to Unix line endings, so that Git stops freaking out every time I switch branches.

See here: <https://github.com/CMSgov/bluebutton-data-server/pull/59/files>.

CBBF-115
CBBF-118
CBBF-121


karlmdavis:
Going to merge this one ahead of review, since there's no functional change (and it's blocking me).

PR #63, opened 2018-01-25
CBBF-100: More script fixes

karlmdavis:
Brian and I just missed these back in the first PR. The problems they cause don't show up on the "happy path", so they weren't caught by our tests, either.

https://issues.hhsdevcloud.us/browse/CBBF-100


karlmdavis:
Then went and fixed/improved some more things in the script while I was in there, because why not?

@dho937 and @glovercloud Can you both please verify that a `mvn clean verify` of this branch works as expected for you? All we need now is some weirdo running Solaris or FreeBSD, and we'll _really_ have all of the platforms covered by our dev team.

PR #68, opened 2018-01-29
CBBF-104: Fixed the patient reference mapping and added code to set t…

glovercloud:
…he EOB status based on the claim final action state.


glovercloud:
He had tests in there already for status and patient reference so I just left it.

glovercloud:
I went around and around on how to hit that switch code path as the tests are not really structured in a way to accommodate different expected results.  I landed on using the existing SAMPLE_U rif data which happens to have the final action set to 'N'(cancelled) in it.  I wasn't sure if the assertMatches was going to work with that data but it appears to pass.  I added a switch within the existing assertEobCommonClaimHeaderData to test both cases.  I believe the patient ref test works as intended as previously written so no change there.

glovercloud:
Also, since there isn't any other claim sample data that I could find other than Carrier with the final action set to 'N', only the Carrier test has an implementation for hitting the new final action code path.

PR #75, opened 2018-02-05
CBBF-135 Map Carrier CARR_LINE_CLIA_LAB_NUM

jcgeer:
- The field CARR_LINE_CLIA_LAB_NUM for Carrier was remapped as a
valueIdentifier from a valueCodeableConcept.

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CarrierClaimTransformer.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CarrierClaimTransformer.java
@@ -6,6 +6,8 @@ import java.util.Optional;
 import org.hl7.fhir.dstu3.model.ExplanationOfBenefit;
 import org.hl7.fhir.dstu3.model.ExplanationOfBenefit.BenefitBalanceComponent;
 import org.hl7.fhir.dstu3.model.ExplanationOfBenefit.ItemComponent;
+import org.hl7.fhir.dstu3.model.Extension;
+import org.hl7.fhir.dstu3.model.Identifier;
 import org.hl7.fhir.dstu3.model.codesystems.BenefitCategory;
 import org.hl7.fhir.dstu3.model.codesystems.ClaimCareteamrole;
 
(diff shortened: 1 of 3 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CarrierClaimTransformer.java
```diff
@@ -187,12 +189,11 @@ private static ExplanationOfBenefit transformClaim(CarrierClaim claimGroup) {
...
-				 * FIXME this should be mapped as an extension valueIdentifier
-				 * instead of as a valueCodeableConcept
-				 */
-				TransformerUtils.addExtensionCoding(item.getLocation(), TransformerConstants.EXTENSION_IDENTIFIER_CCW_CLIA_LAB_NUM,
-						TransformerConstants.EXTENSION_IDENTIFIER_CCW_CLIA_LAB_NUM, claimLine.getCliaLabNumber().get());
+				Extension cliaLabNumberExtension = item.getLocation().addExtension();
+				cliaLabNumberExtension.setUrl(TransformerConstants.EXTENSION_IDENTIFIER_CCW_CLIA_LAB_NUM);
+				cliaLabNumberExtension.setValue(new Identifier()
+								.setSystem(TransformerConstants.EXTENSION_IDENTIFIER_CCW_CLIA_LAB_NUM)
+								.setValue(claimLine.getCliaLabNumber().get()));
```
dho937: Is it possible to create a common method to create these extension Identifiers...similar to addExtensionCoding...create one that says addExtensionIdentifier?  I have a feeling we will be changing other extension codings to extension identifiers.

jcgeer: I had actually mentioned that to Brian when I saw how he had done one of these extension identifiers before (which I used as a reference to perform this change). I think that is a great idea but thought perhaps we might need a separate Jira task? There are multiple places where a common method could be used that I've seen, so I think it'd be best to have a separate ticket for creating the method as well as going through the transformer classes to make the change to use the new method.

dho937: Sounds good.

PR #78, opened 2018-02-06
CBBF-131: Completely reworked Wildfly/JBoss config

karlmdavis:
Goal here is to fix intermittent errors that we're seeing out on the Jenkins server. Specifically, Wildfly was going boom while running the `:reload` command after configuration. No idea why, but it seems to be a race condition in Wildfly itself, as it was intermittent and timing-sensitive.

Instead of using the CLI to configure things, we now just write out a complete `standalone.xml` file for it in one go. This won't really work in production, I think, but is fine for local dev/test, where we don't
have to worry about overwriting someone else's config work.

(While I was at it, I also switched things to use the scripts right out of the `src/main/config/` directory, rather than grabbing the processed versions out of `target/`. Wasn't adding any value -- only complexity.)

https://issues.hhsdevcloud.us/browse/CBBF-131


karlmdavis:
@dho937 Could you checkout this branch and make sure that I didn't break anything on Cygwin? Don't think so, but better safe than sorry.

@dho937, @glovercloud, and @jcgeer also: Looks like one of our earlier PRs broke some test cases. Here's what I'm seeing:

```
Failed tests:
  ExplanationOfBenefitResourceProviderIT.readEobForExistingHHAClaim:130 expected:<188> but was:<188.00>  
  ExplanationOfBenefitResourceProviderIT.searchForEobsByExistingPatient:392 expected:<188> but was:<188.00>
```

Could one of you fix that ASAP?

glovercloud:
Builds on OSX but I also am seeing the test failures.

glovercloud:
Also unrelated to Karl's changes I am seeing tons of javadoc warnings starting at line 1670 in TransformerUtils.  I think we missed that in a separate PR.

glovercloud:
Do you ever restore the "original" standalone files that you are making copies of?

dho937:
I know about the Javadoc warnings.  Can address next week if we have time.

glovercloud:
Verified tests run without errors after bluebutton-data-model fix.

karlmdavis:
> @glovercloud: Do you ever restore the "original" standalone files that you are making copies of?
Nope. They're just there for development purposes: lets us diff them against the modified version to see what config is actually being changed. Also: when we eventually bump Wildfly/JBoss versions, it'll help with updating the config for that, too.

Any concerns?

glovercloud:
I was just thinking either you should restore them or remove the extra steps.  But if you need it for diffs I guess just leave them.  Also, if you run it twice in a row don't you lose your originals anyways?

PR #79, opened 2018-02-06
CBBF-109: FHIR mapping change (HHA)

jcgeer:
Remapping of unitCount, nationalDrugCodeQualifierCode, and nationalDrugCodeQuantity.

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
@@ -1362,15 +1362,6 @@ public final class TransformerUtils {
 		item.setQuantity(qty);
 
 		if (nationalDrugCodeQualifierCode.isPresent()) {
-			// ItemComponent drugCodeQualifierMod = item
-			// .addModifier(TransformerUtils.createCodeableConcept(TransformerConstants.CODING_CCW_NDC_UNIT,
-			// nationalDrugCodeQualifierCode.get()));
-
-			// TransformerUtils.addExtensionCoding(item.getModifier().get(1).addExtension(),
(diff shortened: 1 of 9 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -1354,23 +1357,17 @@ static ItemComponent mapEobCommonItemRevenue(ItemComponent item, ExplanationOfBe
...
 		item.setQuantity(qty);
 
 		if (nationalDrugCodeQualifierCode.isPresent()) {
 			item.addModifier(TransformerUtils.createCodeableConcept(TransformerConstants.CODING_CCW_NDC_UNIT,
-					nationalDrugCodeQualifierCode.get()));
+					nationalDrugCodeQualifierCode.get())).addExtension();
+
+			TransformerUtils.addExtensionCoding(item.getModifier().get(1).addExtension(),
+					TransformerConstants.CODING_CCW_NDC_QTY, TransformerConstants.CODING_CCW_NDC_QTY,
+					String.valueOf(nationalDrugCodeQuantity.get()));
```
dho937: No need to make a change here but I am going to write up a ticket where we can search for modifiers based on their system code.  That way we aren't hard coding index numbers.   I have gotten burnt when I added a modifier above this type of code and I had to change the subscript.

jcgeer: This code is actually a bit outdated. I ran into the issue you're describing on the hard coding of index numbers as it seemed Inpatient did not have a modifier ahead of this one like all the other types did. If you check my most recent commits you can see what I did to work around this for the time being.
Along the same lines, I did not spend too much time with testing to check the correct modifiers, so you will notice with the common method in the test utils that I am passing in a hard coded index so that each claim transformer type knows which modifier to check against.

PR #80, opened 2018-02-07
CBBF-134 Map Carrier CARR_LINE_ANSTHSA_UNIT_CNT

jcgeer:
- The field CARR_ANSTHSA_UNIT_CNT has been mapped to
item.service.extension only when the value is greater than zero.


dho937:
@jcgeer    @glovercloud Forgot one thing...can you add some Junit code (one line of code) to the RifFilesProcessorTest class when we are mapping new fields?   This class makes sure what we are reading in from the flat files matches our tests.  You don't need to do it for this test but in the future; please do.  Thanks.

PR #81, opened 2018-02-07
CBBF-142: ICD Codes have invalid Coding.system

jcgeer:
- Changes to IcdCode.getFhirSystem() to compare icdVersionCode (char
type) to a character value (versus a string)


karlmdavis:
@jcgeer A few things:

1. In general, please try to delete your branches when you merge your PRs. You can always restore them if you change your mind.
2. Please add a test case for this problem. Maybe a new IcdCodeTest class?
3. Given how badly broken this is, it will be a major change for our end users. Can you please update `dev/api-changelog.md`?

PR #91, opened 2018-02-26
CBBF-97: Normalize README.md's line endings.

karlmdavis:
Looks like it got committed with Windows line endings at some point. Normalizing that so that Git no longer things it's modified for all us Unix folk.

Given our `.gitattributes` file, I didn't think that this could have gone sideways in the first place. My best guess is that the file hadn't been changed since we added the `.gitattributes`, and that my recent change to it caused Git to notice the problem.

I followed the instructions here and it looks like this is the only file in this project with this problem: https://help.github.com/articles/dealing-with-line-endings/


karlmdavis:
Need to merge this as it has me blocked -- unable to rebase things locally. Since it's a completely non-functional change (and I just double-checked that), I'm going to merge before @glovercloud or @jcgeer 's review. Can you both please get that review in ASAP, though, so that we can deploy builds including the change?

PR #95, opened 2018-03-26
cbbf-173 set Observation.value even when hctHgbTestTypeCode is present

dho937:
set Observation.value even when hctHgbTestTypeCode is present and hctHgbTestResult is 0


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -1773,7 +1773,7 @@ static ItemComponent mapEobCommonItemCarrierDME(ItemComponent item, ExplanationO
 					calculateVariableReferenceUrl(CcwCodebookVariable.LINE_HCT_HGB_RSLT_NUM),
 					new Reference(hctHgbObservation));
 			item.addExtension(hctHgbObservationReference);
-		} else if (!hctHgbTestTypeCode.isPresent() && hctHgbTestResult.compareTo(BigDecimal.ZERO) == 0) {
+		} else if (hctHgbTestResult.compareTo(BigDecimal.ZERO) == 0) {
```
karlmdavis: I think we can eliminate all of the `else if` and `else` clauses here, can't we -- just leave it as a single `if`?

PR #96, opened 2018-03-26
cbbf-171 enhance code to allow for optional procedure date

dho937:
Enhanced code to allow for optional procedure date when procedure code is present.  Procedure date was assumed to be present when procedure code was also present before; now the date isn't required.

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CCWProcedure.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CCWProcedure.java
@@ -31,11 +31,7 @@ final class CCWProcedure extends IcdCode {
 		Objects.requireNonNull(icdVersionCode);
 		Objects.requireNonNull(procedureDate);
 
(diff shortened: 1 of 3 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CCWProcedure.java
```diff
@@ -31,14 +31,18 @@ public CCWProcedure(Optional<String> icdCode, Optional<Character> icdVersionCode
 		Objects.requireNonNull(icdVersionCode);
 		Objects.requireNonNull(procedureDate);
 
-		this.procedureDate = procedureDate.get();
-
+		if (procedureDate.isPresent()) {
+			this.procedureDate = procedureDate;
+		} else {
+			this.procedureDate = Optional.empty();
+		}
```
karlmdavis: This should just be:

    this.procedureDate = procedureDate;

No `if` checks necessary.

karlmdavis:
Actually, also: can you add a regression test case to cover this?

PR #97, opened 2018-03-28
CBBI-357: Slow DB Queries Bandaid #2

karlmdavis:
Switched to a workaround that allows us to leave all subqueries enabled. It's still gross, but definitely an improvement.

(Work continues to try and determine the actual root cause. Right now it's looking like Postgres is sad that it doesn't have more memory to keep the indexes hot.)

https://issues.hhsdevcloud.us/browse/CBBI-357


karlmdavis:
Note: I tested this out with out prod data and it results in EOB queries that take around 6s. Not great, but WAY better than 200 minutes.

PR #101, opened 2018-05-05
BLUEBUTTON-185: Also search BeneficiaryHistory records for HICNs

karlmdavis:
Allows the Data Server to retrieve beneficiaries whose HICNs have changed, even when searching with an old HICN (provided that the CCW captured that old HICN).

https://jira.cms.gov/browse/BLUEBUTTON-185

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
@@ -1,11 +1,14 @@
 package gov.hhs.cms.bluebutton.server.app.stu3.providers;
 
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Set;
 
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
+import javax.persistence.NonUniqueResultException;
 import javax.persistence.PersistenceContext;
 import javax.persistence.criteria.CriteriaBuilder;
 import javax.persistence.criteria.CriteriaQuery;
@@ -27,6 +30,8 @@ import ca.uhn.fhir.rest.server.IResourceProvider;
 import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import ca.uhn.fhir.rest.server.exceptions.ResourceNotFoundException;
 import gov.hhs.cms.bluebutton.data.model.rif.Beneficiary;
+import gov.hhs.cms.bluebutton.data.model.rif.BeneficiaryHistory;
+import gov.hhs.cms.bluebutton.data.model.rif.BeneficiaryHistory_;
 import gov.hhs.cms.bluebutton.data.model.rif.Beneficiary_;
 
 /**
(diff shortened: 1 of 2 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -199,14 +204,50 @@ private Patient queryDatabaseByHicnHash(String hicnHash) {
...
-		criteria.where(builder.equal(root.get(Beneficiary_.hicn), hicnHash));
+		CriteriaBuilder builder = entityManager.getCriteriaBuilder();
+		Set<String> matchingBeneficiaryIds = new HashSet<>();
+
+		// Search the Beneficiary records for HICN matches.
+		CriteriaQuery<String> beneHicnQuery = builder.createQuery(String.class);
+		Root<Beneficiary> beneHicnQueryRoot = beneHicnQuery.from(Beneficiary.class);
+		beneHicnQuery.select(beneHicnQueryRoot.get(Beneficiary_.beneficiaryId));
+		beneHicnQuery.where(builder.equal(beneHicnQueryRoot.get(Beneficiary_.hicn), hicnHash));
+		matchingBeneficiaryIds.addAll(entityManager.createQuery(beneHicnQuery).getResultList());
```
whytheplatypus: If a match is found here, this could exit early.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -199,14 +204,50 @@ private Patient queryDatabaseByHicnHash(String hicnHash) {
...
+		beneHistoryHicnQuery.where(builder.equal(beneHistoryHicnQueryRoot.get(BeneficiaryHistory_.hicn), hicnHash));
+		matchingBeneficiaryIds.addAll(entityManager.createQuery(beneHistoryHicnQuery).getResultList());
+
+		/*
+		 * Because data is always dirty, we watch out for and throw an error if no
+		 * matches are found or if more than one match is found.
+		 */
+		if (matchingBeneficiaryIds.size() <= 0) {
+			throw new NoResultException();
+		} else if (matchingBeneficiaryIds.size() > 1) {
```
whytheplatypus: If a match is found in both the `BeneficiaryHistory` and the `Beneficiary` sets what prevents this from triggering?

karlmdavis: Nothing: we want it to. From our perspective: if more than one match is found for a given HICN, that means that two different beneficiaries have been assigned the same HICN at various points in time.

What's the right **and secure** thing to do in that case? Beats me, so I'd rather go boom than risk sending out a beneficiary's PHI to the wrong user.

As an aside: We ran a quick analysis on this phenomenon, and of the 163M HICNs we have, only 47K are associated with >1 beneficiary. So it's a low probability event, but one that we definitely want to guard against.

My takeaway from your questions here though is that some of this reasoning needs to be captured in a code comment. I'll do that.

karlmdavis: :+1:

PR #102, opened 2018-05-14
BLUEBUTTON-200: Fix duplicate EOBs bug

karlmdavis:
Fixed the bug in https://jira.cms.gov/browse/BLUEBUTTON-200. This is/was caused by a pretty surprising (to me, at least) JPA problem, where non-lazy-load queries across multiple entities (e.g. an `INNER JOIN` like we're doing here) cause duplicate top-level entities to be returned by JPA: one for each of the result set rows. In this specific case, we were getting a `FooClaim` instance for every `FooClaimLine` row that was found. This is often referred to as the "N+1" problem with JPA, for reasons that escape me.

While I was in there, I also cleaned up the related code quite a bit, as it'd started to get out of hand.

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ClaimType.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ClaimType.java
@@ -14,6 +14,7 @@ import javax.persistence.metamodel.SingularAttribute;
 
 import org.hl7.fhir.dstu3.model.ExplanationOfBenefit;
 
+import gov.hhs.cms.bluebutton.data.model.rif.Beneficiary;
 import gov.hhs.cms.bluebutton.data.model.rif.CarrierClaim;
 import gov.hhs.cms.bluebutton.data.model.rif.CarrierClaim_;
 import gov.hhs.cms.bluebutton.data.model.rif.DMEClaim;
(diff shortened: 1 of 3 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ClaimType.java
```diff
@@ -36,26 +37,32 @@
  * {@link ExplanationOfBenefitResourceProvider}.
  */
 enum ClaimType {
-	CARRIER(CarrierClaim.class, CarrierClaim_.claimId, CarrierClaimTransformer::transform, CarrierClaim_.lines),
+	CARRIER(CarrierClaim.class, CarrierClaim_.claimId, CarrierClaim_.beneficiaryId, CarrierClaimTransformer::transform,
+			CarrierClaim_.lines),
 	
-	DME(DMEClaim.class, DMEClaim_.claimId, DMEClaimTransformer::transform, DMEClaim_.lines),
+	DME(DMEClaim.class, DMEClaim_.claimId, CarrierClaim_.beneficiaryId, DMEClaimTransformer::transform,
```
jcgeer: Should this be DMEClaim_.beneficiaryId?

karlmdavis: :+1:

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -163,245 +144,58 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
-				root.get(SNFClaim_.beneficiaryId));
+	@SuppressWarnings({ "rawtypes", "unchecked" })
+	private <T> List<T> findClaimTypeByPatient(ClaimType claimType, String patientId) {
+		CriteriaBuilder criteriaBuilder = entityManager.getCriteriaBuilder();
 
-		List<SNFClaim> claimEntities = entityManager.createQuery(criteriaQuery).getResultList();
+		CriteriaQuery criteria = criteriaBuilder.createQuery((Class) claimType.getEntityClass());
+		Root root = criteria.from((Class) claimType.getEntityClass());
+		claimType.getEntityLazyAttributes().stream().forEach(a -> root.fetch((PluralAttribute) a));
+		criteria.select(root).distinct(true);
```
karlmdavis: FYI: the `.distinct(true)` here is the actual fix.

PR #104, opened 2018-05-15
BLUEBUTTON-239: Remove HICN hash from returned EOBs

karlmdavis:
It's still present and used behind the scenes for beneficiary mapping, but it will no longer be sent to applications as part of our EOB responses.

Why? Because having it in there has always made me a bit uncomfortable. I'm not convinced that it's a problem, but I'm also unconvinced that it's safe for us to expose.

https://jira.cms.gov/browse/BLUEBUTTON-239


karlmdavis:
@kellytaylorgov @whytheplatypus and @ekivemark: I added you as reviewers on this one since it's a backwards-incompatible, user-facing change. I think it's a good idea, but want some second opinions on that.

PR #105, opened 2018-05-15
BLUEBUTTON-241: Add metrics for transforms and DB queries

karlmdavis:
The app's behavior in production has me wondering if this will point to some obvious perf optimization targets.

https://jira.cms.gov/browse/BLUEBUTTON-241


karlmdavis:
(Reviewers: Be sure to see the second commit message here for details on what it's doing.)

PR #106, opened 2018-05-18
BLUEBUTTON-130: Create reference links and response based on base path

jcgeer:
Changing the path mapping from /baseDstu3/ to /v1/fhir/ so that the frontend does not need to replace the URL values in the fhir response we are sending to them.

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/BlueButtonServerInitializer.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/BlueButtonServerInitializer.java
@@ -63,7 +63,7 @@ public final class BlueButtonServerInitializer implements WebApplicationInitiali
 		ServletRegistration.Dynamic cxfServletReg = servletContext.addServlet("fhirStu3Servlet", stu3Servlet);
 		cxfServletReg.setLoadOnStartup(1);
 		cxfServletReg.addMapping("/v1/fhir/*");
-		cxfServletReg.addMapping("/baseDstu3/*");
+		// cxfServletReg.addMapping("/baseDstu3/*");
 
 		/*
 		 * Register the MetricRegistry and HealthCheckRegistry into the ServletContext,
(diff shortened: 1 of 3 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/BlueButtonServerInitializer.java
```diff
@@ -62,7 +62,7 @@ public void onStartup(ServletContext servletContext) throws ServletException {
 		BlueButtonStu3Server stu3Servlet = new BlueButtonStu3Server();
 		ServletRegistration.Dynamic cxfServletReg = servletContext.addServlet("fhirStu3Servlet", stu3Servlet);
 		cxfServletReg.setLoadOnStartup(1);
-		cxfServletReg.addMapping("/baseDstu3/*");
+		cxfServletReg.addMapping("/v1/fhir/*");
```
whytheplatypus: this tightly couples the `web-server` and the `data-server`, is it possible to let the `data-server` use it's own version path (`baseDstu3` in this case) and enable this behavior in the serialization of the response?

cmsuser1: Haven't found a way to get HAPI (the framework being used here) to do that pre- or during-serialization.

We could add a servlet `Filter` to accomplish it post-serialization, but that's then a second pass through the output, which is what we're trying to help you avoid in the first place.

Given the limitations of our libraries, I'm personally comfortable with the coupling here, since you guys can't change this base URL without breaking devs' compatibility, anyways. But @jcgeer is leading this one, so it's his call.

karlmdavis: Haven't found a way to get HAPI (the framework being used here) to do that pre- or during-serialization.

We could add a servlet Filter to accomplish it post-serialization, but that's then a second pass through the output, which is what we're trying to help you avoid in the first place.

Given the limitations of our libraries, I'm personally comfortable with the coupling here, since you guys can't change this base URL without breaking devs' compatibility, anyways. But @jcgeer is leading this one, so it's his call.

jcgeer:
@whytheplatypus @karlmdavis I believe this should be good to go once you two sign off on it. It successfully builds locally but Jenkins is having some issues.

PR #107, opened 2018-05-24
initial class to download FDA drug data

dho937:
please review my code


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -0,0 +1,164 @@
...
+	 * 
+	 * @param args
+	 *            <p>
+	 *            The non-JVM command line arguments that the application was
+	 *            launched with. Must include:
+	 *            </p>
+	 *            <ol>
+	 *            <li><code>OUTPUT_DIR</code>: the first (and only) argument for
+	 *            this application must be the already-existing path to write the
+	 *            parsed XML codebooks files out to</li>
```
karlmdavis: Looks like the comment here needs to be updated after copy-paste.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -0,0 +1,164 @@
...
+
+		Path outputPath = Paths.get(args[0]);
+		if (!Files.isDirectory(outputPath)) {
+			System.err.println("OUTPUT_DIR does not exist for FDA NDC download.");
+			System.exit(3);
+		}
+
+		// download FDA NDC file
+		String nationalDrugCodeDownloadableFile = "https://www.accessdata.fda.gov/cder/ndctext.zip";
+		String downloadedNdcZipFile = outputPath.toString() + File.separator + "ndctext.zip";
```
karlmdavis: You should represent file paths as a `Path` object in Java these days.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -0,0 +1,164 @@
...
+		try {
+			// connectionTimeout, readTimeout = 10 seconds
+			FileUtils.copyURLToFile(new URL(nationalDrugCodeDownloadableFile), new File(downloadedNdcZipFile), 10000,
+					10000);
+		} catch (IOException e) {
+			System.err.println("socket timeout-ndc file to download " + nationalDrugCodeDownloadableFile);
+			System.err.println("socket timeout-file to download to " + downloadedNdcZipFile);
```
karlmdavis: Not a big deal, but this won't always be a socket timeout. Error message might be misleading to anyone trying to troubleshoot things in the future, if the problem is something else, instead (e.g. unable to write the file because the disk is full).

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -0,0 +1,164 @@
...
+			e.printStackTrace();
+			System.exit(4);
+		}
+
+		// unzip FDA NDC file
+		unzip(downloadedNdcZipFile, outputPath.toString());
+		Files.move(Paths.get(outputPath.toString() + File.separator + "product.txt"),
```
karlmdavis: Should be:

    Paths.get(outputPath.toString(), "product.txt")

That method will automatically insert the correct separator for you. Also, given an existing path you can use `Path.resolve(...)` to get a child path.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -0,0 +1,164 @@
...
+		CharsetDecoder inDec=Charset.forName("windows-1252").newDecoder()
+			.onMalformedInput(CodingErrorAction.REPORT)
+			.onUnmappableCharacter(CodingErrorAction.REPORT);
+
+		CharsetEncoder outEnc=StandardCharsets.UTF_8.newEncoder()
+			.onMalformedInput(CodingErrorAction.REPORT)
+			.onUnmappableCharacter(CodingErrorAction.REPORT);
+
+		try
+		(FileInputStream is = new FileInputStream(outputPath.toString() + File.separator + "fda_products_cp1252.tsv");
```
karlmdavis: You're using the same path more than once, so you should probably create and (re-)use a reference to a `Path` instance.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -0,0 +1,164 @@
...
+			destDir.mkdir();
+		}
+		ZipInputStream zipIn = new ZipInputStream(new FileInputStream(zipFilePath));
+		ZipEntry entry = zipIn.getNextEntry();
+		// iterates over entries in the zip file
+		while (entry != null) {
+			String filePath = destDirectory + File.separator + entry.getName();
+			if (!entry.isDirectory()) {
+				// if the entry is a file, extracts it
+				extractFile(zipIn, filePath);
```
karlmdavis: You can replace some of this code with the `commons-io` dependency you added: https://stackoverflow.com/a/13912353

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/UnzipUtility.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.zip.ZipInputStream;
+
+/**
+ * This utility extracts files and directories of a standard zip file to a
+ * destination directory.
+ * 
+ * @author www.codejava.net
+ *
+ */
+public class UnzipUtility {
```
karlmdavis: This utility class doesn't look like it's used anywhere. Remove it?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2491,4 +2519,75 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+	 */
+	public static String retrieveFDADrugCodeDisplay(String claimDrugCode) {
+
+		if (claimDrugCode.isEmpty())
+			return null;
+
+		// read the entire NDC file the first time and put in a Map
+		if (ndcProductMap.size() == 0) {
+			readFDADrugCodeFile();
+		}
```
karlmdavis: Race condition here: what if two requests come in at the same time?

Also, this should probably be an assignment -- instead of modifying the global variable as a side effect.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2491,4 +2519,75 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+
+		return null;
+	}
+
+	/**
+	 * Reads ALL the PRODUCTNDC and SUBSTANCENAME fields from the FDA NDC Products
+	 * file which was downloaded during the build process
+	 * 
+	 */
+	public static Map<String, String> readFDADrugCodeFile() {
```
karlmdavis: You modify the `static` `ndcProductMap` variable but also return a reference to it. It's likely to lead to errors in the future, as it's not clear if it's a local or global variable.

Instead, create a local instance and return it. This makes this method nice and functional and side-effect-free.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2491,4 +2519,75 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+		}
+
+		String claimDrugCodeReformatted = claimDrugCode.substring(0, 5) + "-" + claimDrugCode.substring(5, 9);
+	    
+		if (ndcProductMap.containsKey(claimDrugCodeReformatted)) {
+			String ndcSubstanceName = ndcProductMap.get(claimDrugCodeReformatted);
+			return ndcSubstanceName;
+		}
+
+		if (!drugCodeLookupMissingFailures.contains(claimDrugCodeReformatted)) {
```
karlmdavis: You should at least add a comment here explaining why you're doing this, as the early return above is easy to miss.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -116,6 +123,16 @@
 	 */
 	private static final Set<CcwCodebookVariable> codebookLookupDuplicateFailures = new HashSet<>();
 
+	/**
+	 * Stores the PRODUCTNDC and SUBSTANCENAME from the failures.
+	 */
+	private static final Map<String, String> ndcProductMap = new HashMap<String, String>();
```
karlmdavis: I think the comment here is misleading? Doesn't have anything to do with failures, does it?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2491,4 +2519,75 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+
+		// read the entire NDC file the first time and put in a Map
+		if (ndcProductMap.size() == 0) {
+			readFDADrugCodeFile();
+		}
+
+		String claimDrugCodeReformatted = claimDrugCode.substring(0, 5) + "-" + claimDrugCode.substring(5, 9);
+	    
+		if (ndcProductMap.containsKey(claimDrugCodeReformatted)) {
+			String ndcSubstanceName = ndcProductMap.get(claimDrugCodeReformatted);
```
karlmdavis: How did you decide which variable(s) to use as the display value here? I'd suggest working with Cat and myself to figure out the most useful format of this. We really shouldn't change it once it's released, so it's best to be a bit careful.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2491,4 +2519,75 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+		String line = "";
+		try {
+			ndcProductsIn.readLine();
+			while ((line = ndcProductsIn.readLine()) != null) {
+				String ndcProductColumns[] = line.split("\t");
+				String nationalDrugCodeManufacturer = StringUtils
+						.leftPad(ndcProductColumns[1].substring(0, ndcProductColumns[1].indexOf("-")), 5, '0');
+				String nationalDrugCodeIngredient = StringUtils.leftPad(ndcProductColumns[1]
+						.substring(ndcProductColumns[1].indexOf("-") + 1, ndcProductColumns[1].length()), 4, '0');
+				ndcProductMap.put(nationalDrugCodeManufacturer + "-" + nationalDrugCodeIngredient,
```
karlmdavis: Should really use a format string for this, instead.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2491,4 +2519,75 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+						ndcProductColumns[13]);
+			}
+			ndcProductsIn.close();
+		} catch (IOException e) {
+			e.printStackTrace();
```
karlmdavis: This is definitely a bug: you should never swallow/ignore exceptions like this. Always go boom.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -422,26 +422,30 @@ public void searchForEobsHasNoDupes() throws FHIRException {
...
+					String resourceId = searchResultEntry.getResource().getId();
+					if (claimIds.contains(resourceId))
+						Assert.assertFalse(claimIds.contains(resourceId));
+					claimIds.add(resourceId);
+				}
+				if (searchResults.getTotal() > 0)
+					Assert.assertFalse(claimIds.isEmpty());
+			});
+		} catch (Exception e) {
+			e.printStackTrace();
```
karlmdavis: You should not swallow/ignore exceptions like this.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PartDEventTransformerTest.java
```diff
@@ -154,6 +156,15 @@ static void assertMatches(PartDEvent claim, ExplanationOfBenefit eob) throws FHI
 		if (claim.getSubmissionClarificationCode().isPresent())
 			TransformerTestUtils.assertInfoWithCodeEquals(CcwCodebookVariable.SUBMSN_CLR_CD,
 					CcwCodebookVariable.SUBMSN_CLR_CD, claim.getSubmissionClarificationCode(), eob);
+
+		try {
+
+			TransformerTestUtils.assertFDADrugCodeDisplayEquals(claim.getNationalDrugCode());
+		} catch (IOException e) {
+			// TODO Auto-generated catch block
+			e.printStackTrace();
```
karlmdavis: You should not swallow/ignore exceptions like this.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerTestUtils.java
```diff
@@ -1995,4 +2000,12 @@ static void assertPeriodEquals(Optional<LocalDate> expectedStartDate, Optional<L
 		if (expectedEndDate.isPresent())
 			assertDateEquals(expectedEndDate.get(), actualPeriod.getEndElement());
 	}
+
+	/**
+	 * @throws IOException
+	 */
+	static void assertFDADrugCodeDisplayEquals(String nationalDrugCode) throws IOException {
```
karlmdavis: This method seems wrong? It's not actually asserting the equality of anything.

dev/api-changelog.md
```diff
@@ -1,5 +1,15 @@
-# API Changelog
+# API Change
```
karlmdavis: Why this change?

dev/api-changelog.md
```diff
@@ -1,5 +1,15 @@
...
 
+## BLUEBUTTON-146: Display NDC (National Drug Code) Substance Names in EOB
+
+Several changes have been made to these entries:
+
+* The `Coding.display` to the EOB has been added for NDC fields in Part D, Carrier and DME.
+
+* The Substance Name will be displayed in the `Coding.display` for NDC fields.
+
+* The FDA NDC product file will be downloaded and "baked" into the application during the build process.
```
karlmdavis: I'd leave this last line out. _We_ care about how it was implemented, but I don't think end users will.

Or instead, maybe mention the source of the data we're using? That might be useful info for folks if they run into display values that don't match their expectations.

dev/api-changelog.md
dho937: Didn't mean to make this change..changed back to original

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -416,36 +416,33 @@ public void searchForEobsByExistingPatient() throws FHIRException {
 	 * @throws FHIRException
 	 *             (indicates test failure)
 	 */
-	@Test
+	// @Test
 	public void searchForEobsHasNoDupes() throws FHIRException {
 		List<Object> loadedRecords = ServerTestUtils
```
dho937: @karlmdavis   I can't get this JUnit to work now..keep getting a http 500 server error..can't call access method.  I didn't change any code in this method.    It works for sample_a but sample_b does not.  I wonder if this search isn't finding any eob's for a certain patient id and that is why it is throwing the error.   Has the sample_b data changed?   I will keep debugging...  Thanks.

karlmdavis: No, I think the server just sometimes chokes on the amount of data in `SAMPLE_B` and the JVM crashes. I've been seeing this locally, intermittently.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -70,23 +70,21 @@ public static void main(String[] args) throws IOException {
...
 			System.exit(4);
 		}
 
 		// unzip FDA NDC file
-		unzip(downloadedNdcZipFile, outputPath.toString());
-		Files.move(Paths.get(outputPath.toString() + File.separator + "product.txt"),
-				Paths.get(outputPath.toString() + File.separator + "fda_products_cp1252.tsv"), REPLACE_EXISTING);
+		unzip(downloadedNdcZipFile.toString(), outputPath.toString());
+		Files.move(Paths.get(outputPath.toString(), "product.txt"),
+				Paths.get(outputPath.toString(), "fda_products_cp1252.tsv"), REPLACE_EXISTING);
```
karlmdavis: Same for the three `toString()` calls here, except it'll be more like `outputPath.resolve("whateveritis").toFile()` for the last two.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -98,10 +96,11 @@ public static void main(String[] args) throws IOException {
 			.onUnmappableCharacter(CodingErrorAction.REPORT);
 
 		try
-		(FileInputStream is = new FileInputStream(outputPath.toString() + File.separator + "fda_products_cp1252.tsv");
+		(FileInputStream is = new FileInputStream(
+				Paths.get(outputPath.toString(), "fda_products_cp1252.tsv").toString());
```
karlmdavis: `toFile()`

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -98,10 +96,11 @@ public static void main(String[] args) throws IOException {
...
 			 BufferedReader reader=new BufferedReader(new InputStreamReader(is, inDec));
 				FileOutputStream fw = new FileOutputStream(
-						outputPath.toString() + File.separator + "fda_products_utf8.tsv");
+						Paths.get(outputPath.toString(), "fda_products_utf8.tsv").toString());
```
karlmdavis: `toFile()`

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -130,13 +129,13 @@ private static void unzip(String zipFilePath, String destDirectory) throws IOExc
 		ZipEntry entry = zipIn.getNextEntry();
 		// iterates over entries in the zip file
 		while (entry != null) {
-			String filePath = destDirectory + File.separator + entry.getName();
+			Path filePath = Paths.get(destDirectory, entry.getName());
```
karlmdavis: Avoid stringish typing: have the method accept `Path`s and use `somePath.resolve("whateveritis")` (and similar) to traverse down from there.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -130,13 +129,13 @@ private static void unzip(String zipFilePath, String destDirectory) throws IOExc
...
 			if (!entry.isDirectory()) {
 				// if the entry is a file, extracts it
-				extractFile(zipIn, filePath);
+				extractFile(zipIn, filePath.toString());
```
karlmdavis: I'd switch this method to take in a `Path`, rather than having to convert to a `String`.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -130,13 +129,13 @@ private static void unzip(String zipFilePath, String destDirectory) throws IOExc
...
 			} else {
 				// if the entry is a directory, make the directory
-				File dir = new File(filePath);
+				File dir = new File(filePath.toString());
```
karlmdavis: `toFile()`

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -124,9 +124,9 @@
 	private static final Set<CcwCodebookVariable> codebookLookupDuplicateFailures = new HashSet<>();
 
 	/**
-	 * Stores the PRODUCTNDC and SUBSTANCENAME from the failures.
+	 * Stores the PRODUCTNDC and SUBSTANCENAME from the downloaded NDC file.
 	 */
-	private static final Map<String, String> ndcProductMap = new HashMap<String, String>();
+	private static Map<String, String> ndcProductMap = new HashMap<String, String>();
```
karlmdavis: I wouldn't initialize this field to anything at all here (except maybe `null`), as it's just more likely to lead to bugs.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2579,15 +2581,16 @@ public static String retrieveFDADrugCodeDisplay(String claimDrugCode) {
...
 				String nationalDrugCodeIngredient = StringUtils.leftPad(ndcProductColumns[1]
 						.substring(ndcProductColumns[1].indexOf("-") + 1, ndcProductColumns[1].length()), 4, '0');
-				ndcProductMap.put(nationalDrugCodeManufacturer + "-" + nationalDrugCodeIngredient,
+				ndcProductHashMap.put(String.format("%s-%s", nationalDrugCodeManufacturer, nationalDrugCodeIngredient),
 						ndcProductColumns[13]);
 			}
 			ndcProductsIn.close();
 		} catch (IOException e) {
 			e.printStackTrace();
+			System.exit(6);
```
karlmdavis: Well, that's more of a boom than I'd probably suggest, lol. Maybe just wrap and rethrow the exception? For example:

    throw new UncheckedIOException("Unable to read NDC code data.", e);

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
karlmdavis: This is better now, but you still have a race condition -- just a harmless one, now. You should add a comment along the lines of:

    /* There's a race condition here: we may initialize this static field more than once if multiple requests come in at the same time. However, the assignment is atomic, so the race and reinitialization is harmless other than maybe wasting a bit of time. */

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2534,7 +2534,7 @@ public static String retrieveFDADrugCodeDisplay(String claimDrugCode) {
 
 		// read the entire NDC file the first time and put in a Map
 		if (ndcProductMap.size() == 0) {
```
karlmdavis: Per above, check to see if this is `null`, instead.

dho937:
@karlmdavis  Merged data-model and data-pipeline for the NDC branches into Master after they built fine thru Jenkins.  Made the additional changes you reviewed yesterday into the data-server project, pushed up and Jenkins builds fine now.  Going to merge first thing Friday morning  to Master unless I hear differently from you.  Thanks again for your detailed reviews...appreciate it!!

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/FDADrugDataUtilityApp.java
```diff
@@ -70,21 +70,21 @@ public static void main(String[] args) throws IOException {
...
 			System.exit(4);
 		}
 
 		// unzip FDA NDC file
-		unzip(downloadedNdcZipFile.toString(), outputPath.toString());
-		Files.move(Paths.get(outputPath.toString(), "product.txt"),
-				Paths.get(outputPath.toString(), "fda_products_cp1252.tsv"), REPLACE_EXISTING);
+		unzip(downloadedNdcZipFile, outputPath);
+		Files.move(Paths.get(outputPath.resolve("product.txt").toFile().getAbsolutePath()),
+				Paths.get(outputPath.resolve("fda_products_cp1252.tsv").toFile().getAbsolutePath()), REPLACE_EXISTING);
```
karlmdavis: The `Paths.get(...)` calls here aren't needed, I think?

PR #109, opened 2018-06-20
Adding paging to the EOBResourceProvider.

jcgeer:


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -178,8 +204,154 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+		 */
+		Integer pageSize = parseParameterFromRequestDetails(requestDetails, "_count");
+		if (pageSize == null) {
+			pageSize = PAGESIZE;
+		}
+
+		Integer startIndex = parseParameterFromRequestDetails(requestDetails, "startIndex");
+		if (startIndex == null) {
+			startIndex = OFFSET;
+		}
```
karlmdavis: This appears to make paging mandatory, which is not what we want to do:

1. We need to allow the BB2 frontend to keep requesting all of the results until it decides it wants to cutover to having us to do that; if we force that change on our end, they can't transition gracefully.
2. The BCDA project will likely never want to use paging, so we want to leave the option for them to request all the data.

I'd suggest some logic here like the following:

```
PagingArguments pagingArgs = new PagingArguments(requestDetails);

...

private static final class PagingArguments {
  private final Optional<Integer> pageSize;
  private final Optional<Integer> startIndex;

  private PagingArguments(RequestDetails requestDetails) {
    ...
  }

  public boolean isPagingRequested() {
    if (pageSize.isPresent() && startIndex.isPresent())
      return true;
    else if (!pageSize.isPresent() && !startIndex.isPresent())
      return false;
    else
      // It's better to let clients requesting mismatched options know they goofed than to try and guess their intent.
      throw new IllegalArgumentException(String.format("Mismatched paging arguments: pageSize='%s', startIndex='%s'", pageSize, startIndex));
  }

  // getters
```

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -178,8 +204,154 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+		Bundle bundle = new Bundle();
+		for (IBaseResource res : resourceList) {
+			BundleEntryComponent entry = bundle.addEntry();
+			entry.setResource((Resource) res);
+		}
+
+		/*
+		 * Create the paging links and add them to the bundle.
+		 */
+		addPagingLinks(bundle, serverBase, beneficiaryId, startIndex, pageSize, numToReturn, numTotalResults);
```
karlmdavis: And here just guard this with:

    if (pagingArgs.isPagingRequested())
      ...

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -178,8 +204,154 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+	/**
+	 * @param requestDetails
+	 *            the {@link RequestDetails} containing additional parameters for
+	 *            the URL in need of parsing out
+	 * @param parameterToParse
+	 *            the parameter to parse from requestDetails
+	 * @return Returns the parsed parameter as an Integer, null if the parameter is
+	 *         not found.
+	 */
+	private Integer parseParameterFromRequestDetails(RequestDetails requestDetails, String parameterToParse) {
```
karlmdavis: `parseParameterFromRequestDetails` --> `parseIntegerParameter`

(The "RequestDetails" in the method name is already implied by the params' signature, but in Java you generally have to disambiguate return type in cases like this, because you could also have a `parseStringParameter` and you can't overload by return type.)

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -178,8 +204,154 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+		if (startIndex > 0) {
+			int start = Math.max(0, startIndex - pageSize);
+			bundle.addLink(new BundleLinkComponent().setRelation(Bundle.LINK_PREV)
+					.setUrl(createPagingLink(serverBase, beneficiaryId, start, pageSize)));
+		}
+
+		/*
+		 * TODO Create a link to the last page. Currently Bundle.LINK_LAST is not
+		 * supported and it does not appear to like setRelation("last").
+		 */
```
karlmdavis: What links (if any) do we/HAPI already include in our EOB `Bundle`s?

Just make sure you're not doing anything that would remove those. (And in fact, I'd add an IT assertion for that.)

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -222,4 +394,71 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
 		return claims.stream().map(c -> claimType.getTransformer().apply(metricRegistry, c))
 				.collect(Collectors.toList());
 	}
+
+	/**
+	 * This FHIR {@link IBundleProvider} contains all resources matching the search
+	 * sorted by claim ID the by claim type. Returns sublist of those resources,
+	 * useful for paging.
+	 */
+	final static class EobBundleProvider implements IBundleProvider {
```
karlmdavis: Does it still make sense to have this functionality encapsulated here, since -- I don't think -- we're otherwise not really using this class anymore?

Genuinely curious what your thoughts are! To me, it's a bit odd to still see it, but I haven't really thought about it much.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -612,6 +612,33 @@ public static String buildEobId(ClaimType claimType, String claimId) {
...
+	 * @param eob
+	 *            the {@link ExplanationOfBenefit} to extract the id from
+	 * @return the <code>claimId</code> field value (e.g. from
+	 *         {@link CarrierClaim#getClaimId()})
+	 */
+	static String getUnprefixedClaimId(ExplanationOfBenefit eob) {
+		String id = null;
+		for (Identifier i : eob.getIdentifier()) {
+			if (i.getSystem().contains("clm_id") || i.getSystem().contains("pde_id")) {
+				id = i.getValue();
```
karlmdavis: For robustness' sake, I'd make this an early return. Then, instead of returning at the end of the loop, throw a `BadCodeMonkeyException` since an ID was expected but none was found.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -403,6 +406,98 @@ public void searchForEobsByExistingPatient() throws FHIRException {
...
+	 * Verifies that
+	 * {@link ExplanationOfBenefitResourceProvider#findByPatient(ca.uhn.fhir.rest.param.ReferenceParam)}
+	 * works as expected for a {@link Patient} that does exist in the DB, with
+	 * paging.
+	 * 
+	 * @throws FHIRException
+	 *             (indicates test failure)
+	 */
+	@Test
+	public void searchForEobsByExistingPatientWithPaging() throws FHIRException {
```
karlmdavis: Similar to the above on "we want to also still support non-paged requests", add some assertions to the other test case(s) here that verify they aren't paged unexpectedly.

(They'd be failing if our test data had enough data in it, but it doesn't, so.)

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
jcgeer: That was definitely my thought; this class isn't really needed at present. My only reasoning for leaving it was the thought that maybe in the future we'd need something similar as we continue to work on paging (though I'm not sure what changes we'd make that would warrant keeping this).
Right now it'd be just as easy to slice the original eob list to return whatever resources paging requires. For the sake of this PR, I think I'll go ahead and do that, and if we ever do need the EobBundleProvider we've got the commit to go back to.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
jcgeer: I placed the check in the @search method, so if paging is not requested it won't even worry about this specific-to-paging-method.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -146,30 +161,41 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
-	 *            matches for {@link ExplanationOfBenefit}s
-	 * @return Returns a {@link List} of {@link ExplanationOfBenefit}s, which
-	 *         may contain multiple matching resources, or may also be empty.
+	 *            {@link ExplanationOfBenefit#getPatient()} to try and find matches
+	 *            for {@link ExplanationOfBenefit}s
+	 * @param startIndex
+	 * 			  an {@link OptionalParam} for the startIndex (or offset) used to 
+	 * 			  determine pagination
+	 * @return Returns a {@link Bundle} of {@link ExplanationOfBenefit}s, which may
+	 *         contain multiple matching resources, or may also be empty.
```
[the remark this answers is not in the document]
jcgeer: Done.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProviderIT.java
jcgeer: Added assertions to check that there are no additional links for paging when paging is not requested. An additional IT was also added for the case where the page size is larger than the number of results to be returned.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -146,30 +161,41 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
 	 */
 	@Search
-	public List<ExplanationOfBenefit> findByPatient(
-			@RequiredParam(name = ExplanationOfBenefit.SP_PATIENT) ReferenceParam patient) {
+	public Bundle findByPatient(
+			@RequiredParam(name = ExplanationOfBenefit.SP_PATIENT) ReferenceParam patient,
+			@OptionalParam(name = "startIndex") String startIndex,
+			RequestDetails requestDetails) {
+		/*
+		 * startIndex is an optional parameter here because it must be handled in the
```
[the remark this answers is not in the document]
jcgeer: Done.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
jcgeer: New class PagingArguments to handle this.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
jcgeer: Removed EobBundleProvider.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
jcgeer: Done.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
jcgeer: Done. Also added exception handling surrounding the case where an invalid argument is provided.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -178,8 +204,157 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
 		eobs.addAll(transformToEobs(ClaimType.OUTPATIENT, findClaimTypeByPatient(ClaimType.OUTPATIENT, beneficiaryId)));
 		eobs.addAll(transformToEobs(ClaimType.PDE, findClaimTypeByPatient(ClaimType.PDE, beneficiaryId)));
 		eobs.addAll(transformToEobs(ClaimType.SNF, findClaimTypeByPatient(ClaimType.SNF, beneficiaryId)));
-		 
-		return eobs;
+
+		eobs.sort(ExplanationOfBenefitResourceProvider::compareByClaimIdThenClaimType);
```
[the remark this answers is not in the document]
jcgeer: Based on this information, I believe that the sort is stable.

https://docs.oracle.com/javase/8/docs/api/java/util/List.html#sort-java.util.Comparator-

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -178,8 +204,157 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+
+		/*
+		 * Create and return a non paged bundle.
+		 */
+		Bundle bundle = new Bundle();
+		for (IBaseResource res : eobs) {
+			BundleEntryComponent entry = bundle.addEntry();
+			entry.setResource((Resource) res);
+		}
+
```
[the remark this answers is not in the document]
jcgeer: Refactored to look similar to this.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -222,4 +397,82 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
 		return claims.stream().map(c -> claimType.getTransformer().apply(metricRegistry, c))
 				.collect(Collectors.toList());
 	}
+
+	private static final class PagingArguments {
```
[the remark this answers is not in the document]
jcgeer: Done.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -222,4 +397,82 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+			return Optional.empty();
+		}
+
+		/*
+		 * @return Returns true if the pageSize or startIndex is present (i.e. paging is
+		 * requested), false if they are not present, and throws an
+		 * IllegalArgumentException if the arguments are mismatched.
+		 */
+		public boolean isPagingRequested() {
+			if (pageSize.isPresent() || startIndex.isPresent())
```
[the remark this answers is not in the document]
jcgeer: Done.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -222,4 +397,82 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+				// than to try and guess their intent.
+				throw new IllegalArgumentException(String
+						.format("Mismatched paging arguments: pageSize='%s', startIndex='%s'", pageSize, startIndex));
+		}
+
+		/*
+		 * @return Returns the pageSize as an integer. Note: the pageSize must exist at
+		 * this point, otherwise paging would not have been requested.
+		 */
+		public Integer getPageSize() {
```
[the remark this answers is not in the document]
jcgeer: Done.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -178,8 +204,157 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+		Integer startIndex = pagingArgs.getStartIndex();
+		String serverBase = pagingArgs.getServerBase();
+
+		/*
+		 * Determine which resources to return based on the pageSize and startIndex.
+		 */
+		int numToReturn;
+		List<ExplanationOfBenefit> resourceList;
+		int numTotalResults = eobs.size();
+		if (pageSize == null || pageSize.equals(Integer.valueOf(0))) {
```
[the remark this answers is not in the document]
jcgeer: Done, but moved elsewhere as part of some refactoring.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -222,4 +397,82 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
 		return claims.stream().map(c -> claimType.getTransformer().apply(metricRegistry, c))
 				.collect(Collectors.toList());
 	}
+
+	private static final class PagingArguments {
+		private Optional<Integer> pageSize;
+		private Optional<Integer> startIndex;
+		private String serverBase;
```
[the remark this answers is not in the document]
jcgeer: Done.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -612,6 +612,32 @@ public static String buildEobId(ClaimType claimType, String claimId) {
...
+		throw new BadCodeMonkeyException("A claim ID was expected but none was found.");
+	}
+
+	/**
+	 * @param eob
+	 *            the {@link ExplanationOfBenefit} to extract the claim type from
+	 * @return the {@link ClaimType}
+	 */
+	static ClaimType getClaimType(ExplanationOfBenefit eob) {
+		String type = eob.getType().getCoding().get(0).getCode();
```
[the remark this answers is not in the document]
jcgeer: Done.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -178,8 +204,157 @@ public ExplanationOfBenefit read(@IdParam IdType eobId) {
...
+		}
+
+		/*
+		 * Create and return a non paged bundle.
+		 */
+		Bundle bundle = new Bundle();
+		for (IBaseResource res : eobs) {
+			BundleEntryComponent entry = bundle.addEntry();
+			entry.setResource((Resource) res);
+		}
```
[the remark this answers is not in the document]
jcgeer: Examining both the old JSON and our newly created ones, there appear to be no values we are failing to set.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -453,19 +415,19 @@ else if (!pageSize.isPresent() && !startIndex.isPresent())
 		 * @return Returns the pageSize as an integer. Note: the pageSize must exist at
 		 * this point, otherwise paging would not have been requested.
 		 */
-		public Integer getPageSize() {
-			return pageSize.get();
+		public Optional<Integer> getPageSize() {
+			return pageSize;
```
karlmdavis: What if the method worked like this, instead?

```
public int getPageSize() {
	if (!isPagingRequested()) throw new BadCodeMonkeyException();
	return pageSize.get();
}
```

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -453,19 +415,19 @@ else if (!pageSize.isPresent() && !startIndex.isPresent())
...
 		 */
-		public Integer getStartIndex() {
+		public Optional<Integer> getStartIndex() {
 			if (startIndex.isPresent()) {
-				return startIndex.get();
+				return startIndex;
 			}
-			return 0;
+			return Optional.of(0);
 		}
```
karlmdavis: What if the method worked like this, instead?

```
public int getStartIndex() {
	if (!isPagingRequested()) throw new BadCodeMonkeyException();
	if (startIndex.isPresent()) {
		return startIndex.get();
	}
	return 0;
}
```

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -634,7 +635,9 @@ static String getUnprefixedClaimId(ExplanationOfBenefit eob) {
 	 * @return the {@link ClaimType}
 	 */
 	static ClaimType getClaimType(ExplanationOfBenefit eob) {
-		String type = eob.getType().getCoding().get(0).getCode();
+		String type = eob.getType().getCoding().stream()
+				.filter(p -> EnumUtils.isValidEnum(ClaimType.class, p.getCode()))
```
karlmdavis: Should be this .`filter(...)` predicate, instead:

```
.filter(c -> c.getSystem().equals(TransformerConstants.CODING_SYSTEM_BBAPI_EOB_TYPE))
```

PR #112, opened 2018-08-21
Bluebutton 267 refresh api schedule

dho937:


dho937:
@karlmdavis @jcgeer Can you review this PR when you get a chance this week? This Refresh Schedule API ticket had changes in the data-model, data-pipeline and data-server projects. I will send you reminders on the other PR's for this ticket. This PR is not building correctly but it is looking for a class that I moved to the data-model project from this project. Thanks.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2521,6 +2595,144 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
 	}
 
+	/**
+	 * Retrieves the Diagnosis display value from a Diagnosis code look up file
+	 * 
+	 * @param icdCode
+	 *            - Diagnosis code
+	 */
+	public static String retrieveIcdCodeDisplay(String icdCode) {
+
```
jcgeer: It looks like you have a lot of code from your ICD display values pull request also contained in this PR (this being one example). I shouldn't think this will be a problem as the merging should take care of it, however I think it is best to make sure that other PR is merged first.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -90,6 +96,8 @@
 import gov.hhs.cms.bluebutton.data.model.rif.OutpatientClaim;
 import gov.hhs.cms.bluebutton.data.model.rif.OutpatientClaimColumn;
 import gov.hhs.cms.bluebutton.data.model.rif.OutpatientClaimLine;
+import gov.hhs.cms.bluebutton.data.model.rif.RifDataloadHistory;
+import gov.hhs.cms.bluebutton.data.model.rif.RifDataloadHistory_;
```
jcgeer: I'm not quite sure how this process will work so I'm asking here and maybe we can discuss further in person.

I see that you've made changes to the data model project to allow for these changes to work, but is there a way we can test the inclusion of those changes here through the Jenkins build before merging this PR? I don't see a PR in the data model for the changes related to this, isn't that something we'll need to create? I guess I don't fully understand how we manage the dependencies on other projects and how we ensure we are pushing changes to both projects at the same time, so any explanation would be helpful.

jcgeer:
@dho937 I think this PR is looking good. I've left two comments; one as more of an fyi and the other asking about project dependency.

PR #113, opened 2018-08-22
Bluebutton 147 display icd values

dho937:


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2520,6 +2542,144 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+
+	/**
+	 * Reads ALL the ICD codes and display values from the DGNS_CD.txt file
+	 * 
+	 */
+	public static Map<String, String> readIcdCodeFile() {
+
+		Map<String, String> icdDiagnosisMap = new HashMap<String, String>();
+		InputStream icdCodeDisplayStream = Thread.currentThread().getContextClassLoader()
+				.getResourceAsStream("DGNS_CD.txt");
```
karlmdavis: You definitely need some comments here about where this resource came from, how it can be updated, and how often that update needs to happen. Really, I'd move both of the resources into a new `src/main/resources/lookups/` subdirectory and add a `README.md` file there with that information. Then, just add a comment here referencing that file, e.g.:

    /* See src/main/resources/lookups/README.md for info on this resource, e.g. where it came from, its format, and how to update it. */

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2520,6 +2542,144 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+
+		BufferedReader icdCodesIn = null;
+		icdCodesIn = new BufferedReader(new InputStreamReader(icdCodeDisplayStream));
+
+		// We want to extract the ICD Diagnosis codes and display values and put in a
+		// Map
+		// for easy retrieval to get the
+		// display value
+		String line = "";
```
karlmdavis: Could you put this in a multiline comment block so it reflows correctly?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2520,6 +2542,144 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+		try {
+			icdCodesIn.readLine();
+			while ((line = icdCodesIn.readLine()) != null) {
+				String icdColumns[] = line.split("\t");
+				icdDiagnosisMap.put(icdColumns[0], icdColumns[1]);
```
karlmdavis: Add a comment (with an example) explaining what each column contains.

karlmdavis: Are diagnostic codes' case (upper vs. lower vs. mixed) consistent in both the file you're using and in the CCW? If not, you'll end up with spurious failed lookups unless you normalize them in your code.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2520,6 +2542,144 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+		}
+
+		return null;
+	}
+
+	/**
+	 * Reads all the procedure codes and display values from the PRCDR_CD.txt file
+	 * 
+	 */
+	public static Map<String, String> readProcedureCodeFile() {
```
karlmdavis: All the same comments from `readIcdCodeFile()` apply to this method.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2520,6 +2542,144 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+	}
+
+	/**
+	 * Retrieves the Procedure code and display value from a Procedure code look up
+	 * file
+	 * 
+	 * @param procedureCode
+	 *            - Procedure code
+	 */
+	public static String retrieveProcedureCodeDisplay(String procedureCode) {
```
karlmdavis: I think this could be default/package scoped instead of `public`, couldn't it?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
karlmdavis: Are we sure that this file doesn't contain any null/missing/invalid values? I'd probably add some checks here to strip out such possibilities, e.g.:
* empty codes
* empty display values
* codes that don't match the expected format (e.g. -- and I'm completely making this up -- if codes are supposed to be 6 characters long always, discard any that aren't)

Just something to think about. Data like this tends to be dirty, and we don't want to propagate that if we can avoid it.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2520,6 +2542,144 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
...
+		}
+
+		return null;
+	}
+
+	/**
+	 * Reads ALL the ICD codes and display values from the DGNS_CD.txt file
+	 * 
+	 */
+	public static Map<String, String> readIcdCodeFile() {
```
karlmdavis: Does this need to be `public`?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
karlmdavis: If this file isn't sorted, please make it that way, so that changes to it in the future have useful diffs.

dev/api-changelog.md
```diff
@@ -1,5 +1,15 @@
...
 
+## BLUEBUTTON-147: Display ICD and Procedure code displays in EOB
+
+Several changes have been made to these entries:
+
+* The `Coding.display` to the EOB has been added for Diagnostic and Procedure codes in the appropriate claim types.
+
+* A descriptive name will be displayed in the `Coding.display` for Diagnosis/Procedure fields.
+
+* The CCW Diagnosis and Procedure Code/Display files are NOT being downloaded from a site when this application is being built.  These files change once a year and are stored on the GDIT mainframe for now.  We will need to determine how to best update these files (DGNS_CD.txt & PRCDR_CD.txt) once a year.
```
karlmdavis: This last bit is only useful info for _us_, not our users. I'd remove it from here and instead put it in code comments.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2520,6 +2542,144 @@ static void mapHcpcs(ExplanationOfBenefit eob, ItemComponent item, Optional<Char
 		});
 	}
 
+	/**
+	 * Retrieves the Diagnosis display value from a Diagnosis code look up file
+	 * 
+	 * @param icdCode
+	 *            - Diagnosis code
+	 */
+	public static String retrieveIcdCodeDisplay(String icdCode) {
```
karlmdavis: This doesn't need to be `public`, does it?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -126,13 +127,33 @@
...
 
+	/**
+	 * Stores the diagnosis ICD codes and their display values
+	 */
+	private static Map<String, String> icdMap = null;
+
+	/**
+	 * Tracks the diagnosis ICD codes that have already had code lookup failures.
+	 */
+	private static final Set<String> icdLookupMissingFailures = new HashSet<>();
```
karlmdavis: The names here are maybe a bit confusing: aren't the diagnoses and procedures **both** ICD codes?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
karlmdavis: Just like with the NDC codes, can you check with Cat that the display values used here are in the format that would likely be most useful for third-party app devs?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
dho937: Made it public

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
dho937: Yes

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
dho937: Okay

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
dho937: Yes, but wanted to make the procedure codes different

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
dho937: changed code to set incoming code to uppercase

dho937: Done

PR #114, opened 2018-09-17
BLUEBUTTON-421: display Proprietary Name & Substance Name as NDC display

dho937:


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2752,8 +2754,10 @@ public static String retrieveFDADrugCodeDisplay(String claimDrugCode) {
 						.leftPad(ndcProductColumns[1].substring(0, ndcProductColumns[1].indexOf("-")), 5, '0');
 				String nationalDrugCodeIngredient = StringUtils.leftPad(ndcProductColumns[1]
 						.substring(ndcProductColumns[1].indexOf("-") + 1, ndcProductColumns[1].length()), 4, '0');
+				// ndcProductColumns[3] - Proprietary Name
+				// ndcProductColumns[13] - Substance Name
```
jcgeer: Looks good. Just thought I'd ask though if these comments are necessary, or if they could be removed/moved elsewhere.

dho937: Just wanted to let the developers know what each column means in the file that we are using.

PR #121, opened 2018-11-09
BLUEBUTTON-169: Currency expressed as valueMoney should be valueQuantity

jcgeer:
A bug was reported of our EOB resources returning extensions for Primary
Payer Paid Amount field as a valueMoney when the FHIR spec states the
value of an extension cannot be a valueMoney. This changes it to a
valueQuantity.

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CarrierClaimTransformer.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CarrierClaimTransformer.java
@@ -61,7 +61,7 @@ final class CarrierClaimTransformer {
 		TransformerUtils.mapEobType(eob, ClaimType.CARRIER, Optional.of(claimGroup.getNearLineRecordIdCode()), 
 				Optional.of(claimGroup.getClaimTypeCode()));
 
-		TransformerUtils.addAdjudicationTotal(eob, CcwCodebookVariable.PRPAYAMT,
+		TransformerUtils.addPrimaryPayerPaidAmount(eob, CcwCodebookVariable.PRPAYAMT,
 				claimGroup.getPrimaryPayerPaidAmount());
 
 		// Common group level fields between Carrier and DME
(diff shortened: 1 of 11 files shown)


karlmdavis:
Before looking at the code, let's discuss whether we even _want_ to merge this change, since it'll be backwards-incompatible (and thanks for pulling the before-and-after JSON to verify that, @jcgeer !).

And another issue to look at is: this can't be the only place where we're making that mistake, can it? Don't we have tons of `valueMoney` extensions? If we do decide to fix it, we should fix it everywhere.

Did @cgreim and/or @ekivemark have opinions? Personally, I'd suggest that we can't deploy this change without first getting approval from all of our production devs. (Or deploying it as part of a later "v2" version of our API.)

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CarrierClaimTransformer.java
```diff
@@ -61,7 +61,7 @@ private static ExplanationOfBenefit transformClaim(CarrierClaim claimGroup) {
 		TransformerUtils.mapEobType(eob, ClaimType.CARRIER, Optional.of(claimGroup.getNearLineRecordIdCode()), 
 				Optional.of(claimGroup.getClaimTypeCode()));
 
-		TransformerUtils.addAdjudicationTotal(eob, CcwCodebookVariable.PRPAYAMT,
+		TransformerUtils.addValueQuantity(eob, CcwCodebookVariable.PRPAYAMT,
```
karlmdavis: I'm not a fan of how this effectively renames the helps-you-understand-things method name of "`addAdjudicationTotal`" to the generic and not-terribly-helpful "`addValueQuantity`". Similar issue with the test assertion methods.

I know that's just a side effect of how this refactoring happened, but I'd recommend fixing it.

jcgeer: Can do. Would these various amounts still be considered adjudications (in which case I'll put back the old name)? I don't entirely understand how "adjudication" relates to these various field amounts, and haven't seen anything that explains that.

ekivemark:
I am ok with deferring this to a V2 API. We will want to do R4 of FHIR early next year anyway and that would be the natural place to make the change,.

- Mark

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/CarrierClaimTransformer.java
karlmdavis: Yes. "Adjudication" is a dumb word for, "decision on coverage/payment that the insurer (i.e. CMS) made." So it covers deductibles, copays, payments, etc.

jcgeer: Ah, that makes a bit more sense then in that context. I've gone ahead and already made that update anticipating that would be the case. Though it looks like we'll be putting off merging this PR anytime soon.

karlmdavis:
Closing this PR, as the project has now been migrated into a new monorepo: https://github.com/CMSgov/beneficiary-fhir-data

PR #128, opened 2019-02-01
BLUEBUTTON-631: Update HAPI-FHIR version to latest release

jcgeer:
Most notable changes:
-Additional dependencies as HAPI has split the client/server codebase
out of hapi-fhir-base and into two separate modules.
-Multiple package changes due to the reorganization/decoupling mentioned
above.
-Certificate updates.


jcgeer:
Note there are some changes to the endpoint response golden files. The metadata file was expected to change as part of this update, the others have been updated due to merging in the work done by Dave for NPI display codes.

bluebutton-server-app/pom.xml
```diff
@@ -51,18 +63,24 @@
 			<groupId>gov.hhs.cms.bluebutton.data.model</groupId>
 			<artifactId>bluebutton-data-model-rif</artifactId>
 		</dependency>
-
+		
+		<!-- https://mvnrepository.com/artifact/ca.uhn.hapi.fhir/hapi-fhir-structures-dstu3 -->
```
karlmdavis: What are you trying to do/convey with these comments in the POM?

bluebutton-server-app/src/test/resources/endpoint-responses/eobByPatientIdAll.json
```diff
@@ -155,7 +155,8 @@
           "identifier": {
             "system": "http://hl7.org/fhir/sid/us-npi",
             "value": "1497758544"
-          }
+          },
+          "display": "CUMBERLAND COUNTY HOSPITAL SYSTEM, INC"
```
karlmdavis: Can you please split these NPI-related changes into a separate PR? If that's the only reason for the updates to the "approved" responses, it'll be a lot simpler to review this PR without them.

If these are needed, doesn't that mean our build is currently broken?

jcgeer:
It seems I had an old metadata.json file in GitHub to begin with (before adding the workaround method to sort a specific list) and now it's looking like a lot has changed. Really, it's just the way the formatting is working when it comes out of that extra method. This is something I probably should have updated in 637, but didn't think about because it was running correctly.

The actual changes for it are simply the reference URLs changing from `.../Profile/<resource-type>` to `.../StructureDefinition/<resource-type>`.

bluebutton-server-app/pom.xml
```diff
@@ -51,18 +63,24 @@
...
 		<dependency>
-			<!-- At least one "structures" JAR must also be included -->
-			<groupId>ca.uhn.hapi.fhir</groupId>
-			<artifactId>hapi-fhir-structures-dstu3</artifactId>
-			<version>${hapi-fhir.version}</version>
+		    <groupId>ca.uhn.hapi.fhir</groupId>
+		    <artifactId>hapi-fhir-server</artifactId>
+		    <version>3.6.0</version>
+		</dependency>
+		<!-- This dependency includes the client HAPI-FHIR classes -->
```
karlmdavis: I know this is silly, but could you move each of these comments down a line so that they're inside the `<dependency/>` blocks? From experience, they tend to get lost later, otherwise.

bluebutton-server-app/src/test/resources/endpoint-responses/metadata.json
```diff
@@ -1,124 +1,92 @@
 {
-  "resourceType": "CapabilityStatement",
-  "status": "active",
-  "date": "2019-01-16T13:49:30-06:00",
-  "publisher": "Centers for Medicare & Medicaid Services",
-  "kind": "instance",
-  "software": {
-    "name": "Blue Button API: Direct",
-    "version": "1.0.0-SNAPSHOT"
+  "resourceType" : "CapabilityStatement",
```
karlmdavis: Like you said, the changes here appear to be minor.

But we should still give our partners a heads up and ask them to verify that this won't break anything, before we merge. Could you reach out to Ryan Nagle and David Gage with a copy of the new response and the diff and ask them to verify that the changes won't break anything for them ASAP?

jcgeer: Looks like you beat me to it.

jcgeer: From Ryan Nagle:
>it shouldn't cause any trouble for us since, as far as i know, we're building a capability statement based on the backend's stated capabilities, but not doing so dynamically. as long as the core operations are the same (e.g., EOB, Patient and Coverage resources), we should be ok

whytheplatypus: looks fine to me

PR #129, opened 2019-02-26
BLUEBUTTON-765: Changes to EOB paging

jcgeer:
- Adding a "first" link to the paged responses
- Fix logic surrounding the "last" link so that startIndex + _count
being greater than the total does not error when navigating to "last"
page.
- Handle the case where "_count=0". Special case per FHIR specs to
return total info.


karlmdavis:
@whytheplatypus What should happen if someone requests `startIndex=100` if there are only 10 actual results; i.e. requests past the end of data?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -211,12 +206,11 @@ public Bundle findByPatient(
...
+			 * FIXME: Due to a bug in HAPI-FHIR described here
+			 * https://github.com/jamesagnew/hapi-fhir/issues/1074 paging for count=0 is not
+			 * working correctly. Review bluebutton-data-server PR #129 for necessary code
+			 * changes when this issue is resolved.
 			 */
-			if (pagingArgs.getPageSize() == 0) {
-				throw new InvalidRequestException("Invalid request - the page size should not be zero.");
-			}
-
 			int numToReturn = Math.min(pagingArgs.getPageSize(), eobs.size());
```
karlmdavis:
```suggestion
			int endIndex = Math.min(pagingArgs.getStartIndex() + pagingArgs.getPageSize(), eobs.size());
```

This won't apply cleanly: you'll have to update the next line as well to accommodate it. Apologies.

My real concern, though, is that it _looks_ like this code was buggy but none of the tests caught it. I think Step 1 is to write a test case that proves/disproves the bug.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -289,19 +283,29 @@ private void addPagingLinks(Bundle bundle, PagingArguments pagingArgs, String be
...
 					.setUrl(createPagingLink(serverBase, beneficiaryId, startIndex + pageSize, pageSize)));
-
-			int start = (numTotalResults / pageSize - 1) * pageSize;
-			bundle.addLink(new BundleLinkComponent().setRelation("last")
-					.setUrl(createPagingLink(serverBase, beneficiaryId, start, pageSize)));
 		}
 
-		if (startIndex > 0) {
-			int start = Math.max(0, startIndex - pageSize);
+		if (startIndex - pageSize >= 0) {
```
karlmdavis: Just to avoid inviting bugs unnecessarily, I'd suggest removing the guard clause here; always include "first" links.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -289,19 +283,29 @@ private void addPagingLinks(Bundle bundle, PagingArguments pagingArgs, String be
...
-					.setUrl(createPagingLink(serverBase, beneficiaryId, start, pageSize)));
+					.setUrl(createPagingLink(serverBase, beneficiaryId, startIndex - pageSize, pageSize)));
+		}
+
+		/*
+		 * This formula rounds numTotalResults down to the nearest multiple of pageSize
+		 * that's less than and not equal to numTotalResults
+		 */
+		int lastIndex = (numTotalResults - 1) / pageSize * pageSize;
+		if (startIndex < lastIndex) {
```
karlmdavis: Just to avoid inviting bugs unnecessarily, I'd suggest removing the guard clause here; always include "last" links.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -289,20 +282,26 @@ private void addPagingLinks(Bundle bundle, PagingArguments pagingArgs, String be
...
 			bundle.addLink(new BundleLinkComponent().setRelation(Bundle.LINK_PREV)
-					.setUrl(createPagingLink(serverBase, beneficiaryId, start, pageSize)));
+					.setUrl(createPagingLink(serverBase, beneficiaryId, startIndex - pageSize, pageSize)));
 		}
+
+		/*
+		 * This formula rounds numTotalResults down to the nearest multiple of pageSize
+		 * that's less than and not equal to numTotalResults
+		 */
+		int lastIndex = (numTotalResults - 1) / pageSize * pageSize;
```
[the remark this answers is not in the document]
jcgeer: My initial idea for `_count=0` was to not even go into the methods for creating paging links at all. Essentially skipping the lines where we would add resources/links to the bundle and going straight to adding the total information to the bundle.

I suppose though the issue with that would be we're not adding the "last" and "first" links for `_count=0` then, and while that won't necessarily cause us issues it might be confusing to someone to see us not including them for this one specific scenario. If we still want to include those links for this case as well, I suppose the alternative would be an if statement setting `lastIndex = numTotalResults - 1` if the page size is 0.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProviderIT.java
```diff
@@ -541,6 +540,43 @@ public void searchForEobsByExistingPatientWithPaging() throws FHIRException {
...
+		List<Object> loadedRecords = ServerTestUtils
+				.loadData(Arrays.asList(StaticRifResourceGroup.SAMPLE_A.getResources()));
+		IGenericClient fhirClient = ServerTestUtils.createFhirClient();
+
+		Beneficiary beneficiary = loadedRecords.stream().filter(r -> r instanceof Beneficiary).map(r -> (Beneficiary) r)
+				.findFirst().get();
+
+		Bundle searchResults = fhirClient.search().forResource(ExplanationOfBenefit.class)
+				.where(ExplanationOfBenefit.PATIENT.hasId(TransformerUtils.buildPatientId(beneficiary))).count(3)
+				.returnBundle(Bundle.class).execute();
```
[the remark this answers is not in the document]
jcgeer: I agree with you on this, but am not sure that's the right solution. If we add that assert then we'll need to tweak the count and that assertion anytime we add a new claim type.

I'm thinking rather, what if I changed the other method name `searchForEobsByExistingPatientWithEvenPaging`, so we have both odd and even, and I include in the comments for it how it too is verifying we don't run into the out of bounds exception.

Thoughts?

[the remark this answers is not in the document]
karlmdavis: 👍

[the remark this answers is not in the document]
karlmdavis: Well, sort of. The reason this feels a bit off is that "even" and "odd" aren't quite what you're checking -- you're checking whether or not the total count is evenly divisible by the page size. Not sure how to record it to capture that nuance, though.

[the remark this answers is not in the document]
jcgeer: Understood. I've updated the comments and method name for now but can think on a way to better explain that.

dho937:
@jcgeer Is there a certain piece you want me to review?

jcgeer:
> @jcgeer Is there a certain piece you want me to review?

Specifically the following two commits:
https://github.com/CMSgov/bluebutton-data-server/pull/129/commits/7a0e4734c220188f962366e755adab1d0e1cea13
https://github.com/CMSgov/bluebutton-data-server/pull/129/commits/5ed2cc6da5786aa726fda51d804b26781f53e754

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -211,15 +206,13 @@ public Bundle findByPatient(
 		PagingArguments pagingArgs = new PagingArguments(requestDetails);
 		if (pagingArgs.isPagingRequested()) {
 			/*
-			 * A page size of 0 is odd enough that we should throw an exception.
+			 * FIXME: Due to a bug in HAPI-FHIR described here
+			 * https://github.com/jamesagnew/hapi-fhir/issues/1074 paging for count=0 is not
+			 * working correctly. Review bluebutton-data-server PR #129 for necessary code
+			 * changes when this issue is resolved.
```
karlmdavis: @jcgeer As we discussed in HipChat, I think you need to add some mention of how `startIndex < 0` and `_count <=0` edge cases are handled in the JavaDoc for those method parameters.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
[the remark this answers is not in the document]
karlmdavis: Just for future-proofing concerns, I'd suggest fixing this code so that it won't divide by zero or do other goofy things if the HAPI issues with out-of-bounds `startIndex` and `_count` are fixed.

bluebutton-server-app/src/test/resources/endpoint-responses/eobByPatientIdPaged.json
```diff
@@ -1,15 +1,23 @@
 {
   "resourceType": "Bundle",
-  "id": "ed3c3389-92c5-4927-9c90-8cba20e721d2",
+  "id": "09c1904a-e684-4acf-a5d3-5de113feeab9",
```
karlmdavis: Can you remove/revert the spurious time, ID, and port changes from this PR? I did that on my other SAMHSA PR manually and A) wasn't too much work, and B) made the diff much cleaner.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -289,20 +282,31 @@ private void addPagingLinks(Bundle bundle, PagingArguments pagingArgs, String be
...
+
+		/*
+		 * This formula rounds numTotalResults down to the nearest multiple of pageSize
+		 * that's less than and not equal to numTotalResults
+		 */
+		int lastIndex;
+		try {
+			lastIndex = (numTotalResults - 1) / pageSize * pageSize;
+		} catch (ArithmeticException e) {
+			throw new ArithmeticException(String.format("Cannot divide by zero: pageSize=%s", pageSize));
```
karlmdavis: I'm confused... why catch an `ArithmeticException` and just throw a new one? Why not instead, just ensure an empty result is returned or return a `400`?

jcgeer: That's my bad; I meant to catch the ArithmeticException and return an InvalidRequestException (http 400) but in typing it out clearly didn't accomplish that.

karlmdavis: lol

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -381,4 +382,21 @@ private String createPagingLink(String theServerBase, String patientId, int star
...
+	 *            claims) to filter
+	 */
+	private void filterSamhsa(List<ExplanationOfBenefit> eobs) {
+		ListIterator<ExplanationOfBenefit> eobsIter = eobs.listIterator();
+		while (eobsIter.hasNext()) {
+			ExplanationOfBenefit eob = eobsIter.next();
+			if (samhsaMatcher.test(eob))
+				eobsIter.remove();
+		}
+	}
```
karlmdavis: Pretty sure this isn't supposed to be here, right? Looks like an "oops" merge?

jcgeer: I think you're right. Will check and remove.

jcgeer: I think this is supposed to be here. It looks to me like BLUEBUTTON-786 has been merged into master with these changes.
https://github.com/CMSgov/bluebutton-data-server/pull/131/files#diff-13d1fb4e2e1d500e4a24888d773f166b

karlmdavis: Yeah, but why would that also be showing up in this PR? Everything else that came in from your merge to master didn't...

jcgeer: Ah, good point. I am not sure why that's showing as a change here...

karlmdavis: I think you can just go remove it in a new commit. Keeps the history a bit cleaner, that way (`git blame` won't list you as the culprit for those additions).

jcgeer: Right. Will do.

jcgeer:
> Oof, now it's even more goofed. See this: https://github.com/CMSgov/bluebutton-data-server/pull/129/files
> 
> I'd suggest taking a step back and figuring out how best to proceed. You _might_ want to start a new brach and `git cherry-pick` the (verified clean) commits that you want into it.
> 
> Sorry: sometimes Git kinda' sucks.

jcgeer:
@karlmdavis Just finished creating this new PR for it.
https://github.com/CMSgov/bluebutton-data-server/pull/138

jcgeer:
Closing this PR as it has been replaced by https://github.com/CMSgov/bluebutton-data-server/pull/138

PR #130, opened 2019-03-05
Bluebutton 754 refactor deprecated hapi fhir code

jcgeer:


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -679,7 +679,7 @@ public static IdDt buildPatientId(String beneficiaryId) {
 	 *            the {@link Beneficiary} to compute a {@link Coverage#getId()} for
 	 * @return the {@link Coverage#getId()} value to use for the specified values
 	 */
-	static IdDt buildCoverageId(MedicareSegment medicareSegment, Beneficiary beneficiary) {
+	public static IdDt buildCoverageId(MedicareSegment medicareSegment, Beneficiary beneficiary) {
```
karlmdavis: Why'd this get marked `public`? No big deal, really, but I don't see any new uses of it in the rest of the changeset.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/ServerCapabilityStatementIT.java
```diff
@@ -39,7 +39,7 @@
 	public void getCapabilities() {
 		IGenericClient fhirClient = ServerTestUtils.createFhirClient();
 
-		CapabilityStatement capabilities = fhirClient.fetchConformance().ofType(CapabilityStatement.class).execute();
+		CapabilityStatement capabilities = fhirClient.capabilities().ofType(CapabilityStatement.class).execute();
```
karlmdavis: Just curious: is the `.ofType(CapabilityStatement.class)` still required on these calls?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
jcgeer: You're right; there aren't any new uses. But something seems to have changed where it is no longer visible in the EndpointComparatorIT unless I mark it public. Not sure if it's just something locally on my end. Will look into it more.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/ServerCapabilityStatementIT.java
jcgeer: The new method is still returning as `IFetchConformanceUntyped`, so it looks like we still need it.

PR #131, opened 2019-03-06
BLUEBUTTON-786: Initial design & implementation for SAMHSA filtering (Carrier & DME support)

karlmdavis:
Currently only fully supports Carrier and DME claims -- all other claim types will be completely suppressed when the option is enabled.

https://jira.cms.gov/browse/BLUEBUTTON-786


dev/design-samhsa-filtering.md
```diff
@@ -0,0 +1,113 @@
...
+|Outpatient|ICD Diagnosis|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|Outpatient|ICD Procedure|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|SNF|DRG|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|SNF|CPT|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|SNF|ICD Diagnosis|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|SNF|ICD Procedure|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|Part D Events|DRG|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|Part D Events|CPT|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|Part D Events|ICD Diagnosis|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
+|Part D Events|ICD Procedure|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|_(TODO)_|
```
karlmdavis: The table sucks to read in Markdown format. See here for GH's more user friendly rendering of this doc: <https://github.com/CMSgov/bluebutton-data-server/blob/bluebutton-786-samhsa-filtering-carrier-dme/dev/design-samhsa-filtering.md>

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/SamhsaMatcher.java
```diff
@@ -0,0 +1,412 @@
...
+import com.justdavis.karl.misc.exceptions.unchecked.UncheckedIoException;
+
+/**
+ * A {@link Predicate} that, when <code>true</code>, indicates that an
+ * {@link ExplanationOfBenefit} (i.e. claim) is SAMHSA-related.
+ *
+ * See <code>/bluebutton-data-server.git/dev/design-samhsa-filtering.md</code>
+ * for details on the design of this feature.
+ */
+@Component
```
karlmdavis: Minor bug: for performance reasons, this should be bound/configured as a singleton. Accordingly, it should also be marked as thread safe (it should already be; just need to document that).

This... probably doesn't matter, because I _think_ `ExplanationOfBenefitResourceProvider` is a singleton (and it's the only bean that uses it)? But, either way, it's best not to leave that unspecified -- really don't want to be parsing 6 CSV files on every request.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/SamhsaMatcher.java
```diff
@@ -0,0 +1,412 @@
...
+		icd10DiagnosisCode = icd10DiagnosisCode.trim();
+		icd10DiagnosisCode = icd10DiagnosisCode.replaceFirst("\\.", "");
+		icd10DiagnosisCode = icd10DiagnosisCode.toUpperCase();
+
+		return icd10DiagnosisCode;
+	}
+
+	/**
+	 * @param diagnoses
+	 *            the procedure {@link CodeableConcept} to check
```
jcgeer: For conformity's sake, you might change this comment to `@param procedureConcept` to match the parameter name the method is actually using.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/SamhsaMatcher.java
```diff
@@ -0,0 +1,412 @@
...
+		return false;
+	}
+
+	/**
+	 * @param procedureCoding
+	 *            the procedure {@link Coding} to check
+	 * @return <code>true</code> if the specified procedure {@link Coding} matches
+	 *         one of the {@link #icd9DiagnosisCodes} entries, <code>false</code> if
+	 *         it does not
+	 */
```
jcgeer: This comment is misleading, as I believe you want `{@link #cptCodes}`.

dev/design-samhsa-filtering.md
```diff
@@ -0,0 +1,113 @@
...
+
+> In accordance with 42 U.S.C. 290dd-2 and the implementing regulations at 42 CFR part 2, CMS does not share beneficiary identifiable claims data relating to the diagnosis and treatment of alcohol and substance abuse without the explicit written consent of the beneficiary.
+
+## Filtering Approach
+
+In order to determine which claims are SAMHSA-related, CMS has previously identified various diagnosis and procedure codes as being indicative of SAMHSA-related treatment. Any Part A or Part B claims referencing those codes are thus suppressed, which accomplishes the required SAMHSA filtering. This project takes the same approach:
+
+1. The `/bluebutton-server-app/src/main/resources/samhsa-related-codes/codes-*` CSV files store the designated SAMHSA-related codes.
```
jcgeer: How were these files created? And if more codes are added/identified later as being indicative of SAMHSA-related treatment, how will they be updated?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/SamhsaMatcher.java
```diff
@@ -0,0 +1,412 @@
...
+
+	/**
+	 * Constructs a new {@link SamhsaMatcher}, loading the lists of SAMHSA-related
+	 * codes from the classpath.
+	 */
+	public SamhsaMatcher() {
+		this.drgCodes = Collections
+				.unmodifiableList(resourceCsvColumnToList("samhsa-related-codes/codes-drg.csv", "MS-DRGs"));
+		this.cptCodes = Collections
+				.unmodifiableList(resourceCsvColumnToList("samhsa-related-codes/codes-cpt.csv", "CPT Code"));
```
karlmdavis: Should probably be trimming whitespace off of these, at least.

dho937:
@karlmdavis   @jcgeer @cthayes Intense work on this!!  This is complex and only going to get more complex as we deal with the other claim types.   One thought I had though after reviewing the code is that there are a lot of methods to check if a code is a SAMHSA code solely because of the all of the different FHIR formats (i.e. Coding, CodeableConect, Diagnosis, Procedure, and others..).   

Couldn't another option to attack this would be to check for these codes right after they are read from the database and prior to the DataTransformer code?   This would eliminate the need to deal with all the different FHIR types we have chosen to map our data to.  The data obviously would be in its raw format (mostly VARCHAR data) and maybe there could be less isSAMHSA??? and/or containsSAMHSA??? methods...  

Just food for thought...otherwise good stuff Karl!!

cthayes:
@karlmdavis This has a merge conflict now.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/SamhsaMatcher.java
karlmdavis: Beans are singletons by default in Spring: https://stackoverflow.com/a/42599882/1851299

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/SamhsaMatcher.java
karlmdavis: Yup! Thanks!

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/SamhsaMatcher.java
karlmdavis: Yup! Thanks!

dev/design-samhsa-filtering.md
karlmdavis: Great point! I'll update the design doc with details.

karlmdavis:
@dho937 Yeah, I thought for a while about whether to do the filtering pre- or post-transform. I ended up coming down on the side of post-transform, as I didn't want to have to deal with the fifty separate columns for diagnosis and procedure codes things that we have going on in our RIF. At least with the FHIR, everything's collected into `List`s already.

Just a judgement call, really; both approaches have pros & cons, I think.

bluebutton-server-app/src/test/resources/endpoint-responses/metadata.json
```diff
@@ -46,6 +46,9 @@
         "name" : "patient",
         "type" : "reference",
         "documentation" : "The reference to the patient"
+      }, {
+        "name" : "excludeSAMHSA",
+        "type" : "string"
```
karlmdavis: @whytheplatypus Added you as a review here, given this change to our `/v1/fhir/metadata` response. Any concerns?

(I'd suggest that you should strip this out from the frontend's `/metadata` response, if it would otherwise come through. But I wouldn't think that should block us merging this change; doesn't seem like a big deal if it's goofy for a week-ish.)

karlmdavis: @Gage punted to @ekivemark, who had this to say on HipChat:

> I agree, I don’t think it causes problems but we can add a story to our backlog to enable us to filter the value out of the capability statement. We already hack the statement to add smart on fhir info.

ekivemark: @karlmdavis  Can we get a copy of the new JSON CapabilityStatement. It will be helpful for setting up a test to make sure we suppress excludeSAMSHA correctly.

PR #132, opened 2019-03-06
BLUEBUTTON-786: Minor bugfixes and tweaks to NDC downloader

karlmdavis:
Bugfixes:

* Add `<pluginExecution/>` to POM so that it runs in Eclipse.
    * The lack of this was making it unpossible to run unit tests in Eclipse except in a full Maven run.
* Do the download, unzipping, etc. in a temp directory so that unecessary files don't land in our WAR.

Other tweaks made while I was there just to tie things together a bit better and speed up the build a tiny bit.

https://jira.cms.gov/browse/BLUEBUTTON-786 (except not really; I just ran into this while I was working on it)


karlmdavis:
@dho937 @jcgeer @cthayes Could I get a review on this when you have a chance? Not urgent, but I'd like to start cutting down on the number of open PRs we have lingering...

PR #133, opened 2019-03-12
BLUEBUTTON-814: Update endpoint response files for NPI changes

jcgeer:
With the NPI changes brought in the golden files need to be updated to
reflect the changes in the endpoint response files.


jcgeer:
> Looks fine, though I'm surprised that none of our other doc types are here -- I'd guess they just don't have valid NPIs?

That appears to be the case. I'd have to review all of Dave's previous PR again to confirm myself, but with his changes in place these are the only files updated, so I'm guessing it's accurate.

dho937:
@karlmdavis @jcgeer   That is correct Karl.  When I made the NPI changes; I only put in a couple valid NPI codes in a couple record types.  I didn't do this for all the record types.

PR #134, opened 2019-03-14
BLUEBUTTON-767: Add paging to the Patient resource

jcgeer:
- Adding paging to the Patient resource provider as described in the
FHIR spec (see https://www.hl7.org/fhir/search.html ).
- Pulled out PagingArguments into its own class to reduce code
duplication.


jcgeer:
Commit history is getting ugly with the merges to get it up to date with master, created a new PR here https://github.com/CMSgov/bluebutton-data-server/pull/139

PR #139, opened 2019-04-03
BLUEBUTTON-767: Add paging to the Patient resource

jcgeer:
- Adding paging to the Patient resource provider as described in the
FHIR spec (see https://www.hl7.org/fhir/search.html ).
- Moved link creation/adding to the bundle into the PagingArguments
class

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
@@ -218,25 +218,9 @@ public final class ExplanationOfBenefitResourceProvider implements IResourceProv
 
 		eobs.sort(ExplanationOfBenefitResourceProvider::compareByClaimIdThenClaimType);
 
-		Bundle bundle = new Bundle();
 		PagingArguments pagingArgs = new PagingArguments(requestDetails);
-		if (pagingArgs.isPagingRequested()) {
-			/*
-			 * FIXME: Due to a bug in HAPI-FHIR described here
-			 * https://github.com/jamesagnew/hapi-fhir/issues/1074 paging for count=0 is not
-			 * working correctly. Review bluebutton-data-server PR #129 for necessary code
-			 * changes when this issue is resolved.
-			 */
-			int endIndex = Math.min(pagingArgs.getStartIndex() + pagingArgs.getPageSize(), eobs.size());
-			List<IBaseResource> resources = eobs.subList(pagingArgs.getStartIndex(), endIndex);
-			bundle = TransformerUtils.addResourcesToBundle(bundle, resources);
-			pagingArgs.addPagingLinks(bundle, "/ExplanationOfBenefit?", ExplanationOfBenefit.SP_PATIENT, beneficiaryId, eobs.size());
-		} else {
-			bundle = TransformerUtils.addResourcesToBundle(bundle, eobs);
-		}
-
-		bundle.setTotal(eobs.size());
-
+		Bundle bundle = TransformerUtils.createBundle(pagingArgs, "/ExplanationOfBenefit?",
+				ExplanationOfBenefit.SP_PATIENT, beneficiaryId, eobs);
 		return bundle;
 	}
 
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
@@ -19,10 +19,8 @@ import javax.persistence.metamodel.PluralAttribute;
 import javax.persistence.metamodel.SingularAttribute;
 
 import org.hl7.fhir.dstu3.model.Bundle;
-import org.hl7.fhir.dstu3.model.Bundle.BundleEntryComponent;
 import org.hl7.fhir.dstu3.model.ExplanationOfBenefit;
 import org.hl7.fhir.dstu3.model.IdType;
-import org.hl7.fhir.dstu3.model.Resource;
 import org.hl7.fhir.instance.model.api.IBaseResource;
 import org.springframework.stereotype.Component;
 
@@ -200,7 +198,7 @@ public final class ExplanationOfBenefitResourceProvider implements IResourceProv
 
 		String beneficiaryId = patient.getIdPart();
 
-		List<ExplanationOfBenefit> eobs = new ArrayList<ExplanationOfBenefit>();
+		List<IBaseResource> eobs = new ArrayList<IBaseResource>();
 		/*
 		 * The way our JPA/SQL schema is setup, we have to run a separate search for
 		 * each claim type, then combine the results. It's not super efficient, but it's
@@ -230,11 +228,11 @@ public final class ExplanationOfBenefitResourceProvider implements IResourceProv
 			 * changes when this issue is resolved.
 			 */
 			int endIndex = Math.min(pagingArgs.getStartIndex() + pagingArgs.getPageSize(), eobs.size());
-			List<ExplanationOfBenefit> resources = eobs.subList(pagingArgs.getStartIndex(), endIndex);
-			bundle = addResourcesToBundle(bundle, resources);
-			pagingArgs.addPagingLinks(bundle, "/ExplanationOfBenefit?", "&patient=", beneficiaryId, eobs.size());
+			List<IBaseResource> resources = eobs.subList(pagingArgs.getStartIndex(), endIndex);
+			bundle = TransformerUtils.addResourcesToBundle(bundle, resources);
+			pagingArgs.addPagingLinks(bundle, "/ExplanationOfBenefit?", ExplanationOfBenefit.SP_PATIENT, beneficiaryId, eobs.size());
 		} else {
-			bundle = addResourcesToBundle(bundle, eobs);
+			bundle = TransformerUtils.addResourcesToBundle(bundle, eobs);
 		}
 
 		bundle.setTotal(eobs.size());
@@ -247,7 +245,7 @@ public final class ExplanationOfBenefitResourceProvider implements IResourceProv
 	 * 
 	 * @param eob2 an {@link ExplanationOfBenefit} to be compared
 	 */
-	private static int compareByClaimIdThenClaimType(ExplanationOfBenefit eob1, ExplanationOfBenefit eob2) {
+	private static int compareByClaimIdThenClaimType(IBaseResource res1, IBaseResource res2) {
 		/*
 		 * In order for paging to be meaningful (and stable), the claims have to be
 		 * consistently sorted across different app server instances (in case page 1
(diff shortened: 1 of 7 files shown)


jcgeer:
@karlmdavis @dho937 @cthayes  When you review this, you'll notice that `PatientResourceProviderIT` doesn't have quite as many tests for the paging as I do in `ExplanationOfBenefitResourceProviderIT`.

This is in part due to the Patient resource only ever having one page, but also because the paging code it is making use of is exactly the same. I didn't think it needed to be tested again just for a different resource (I could be wrong). I can certainly add those additional test cases into here, but wanted to hear your thoughts on it as it seems a bit of a waste to me. Especially considering that there are two different search types (logical id and hicn hash), I'd have duplicated test code between those as well.

Anyway, we can have a discussion on whether or not and how we want to include all those additional tests (perhaps a separate PagingIT to not clutter up the other ITs with paging related tests?), but I thought I'd put this out there first.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -159,11 +172,33 @@ public Patient read(@IdParam IdType patientId) {
...
+			 * working correctly.
+			 */
+			int endIndex = Math.min(pagingArgs.getStartIndex() + pagingArgs.getPageSize(), patients.size());
+			List<Patient> resources = patients.subList(pagingArgs.getStartIndex(),
+					endIndex);
+			bundle = addResourcesToBundle(bundle, resources);
+			pagingArgs.addPagingLinks(bundle, "/Patient", "&_id=", logicalId.getValue(), patients.size());
+		} else {
+			bundle = addResourcesToBundle(bundle, patients);
+		}
```
karlmdavis: It feels like the code there could be pulled out to a helper method. You've got it copy-pasted more than three times now, which is generally a good indication that the abstraction would be worth it.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -276,4 +340,23 @@ private Patient queryDatabaseByHicnHash(String hicnHash) {
 		Patient patient = BeneficiaryTransformer.transform(metricRegistry, beneficiary);
 		return patient;
 	}
+
+	/**
+	 * @param bunlde
```
karlmdavis:
```suggestion
	 * @param bundle
```

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -276,4 +340,23 @@ private Patient queryDatabaseByHicnHash(String hicnHash) {
...
+	 *            a {@link Bundle} to add the list of {@link ExplanationOfBenefit}
+	 *            resources to.
+	 * @param list
```
karlmdavis:
```suggestion
	 * @param patients
```

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProviderIT.java
```diff
@@ -116,6 +147,40 @@ public void searchForExistingPatientByHicnHash() {
...
+		/*
+		 * Verify that only the first and last paging links exist, since there should
+		 * only be one page.
+		 */
+		Assert.assertNotNull(searchResults.getLink("first"));
+		Assert.assertNull(searchResults.getLink(Bundle.LINK_NEXT));
+		Assert.assertNull(searchResults.getLink(Bundle.LINK_PREV));
+		Assert.assertNotNull(searchResults.getLink("last"));
+	}
+
```
karlmdavis: I'd also update the "without paging" tests to verify that they _don't_ add any paging links.

bluebutton-server-app/src/test/resources/endpoint-responses/metadata.json
```diff
@@ -79,7 +79,12 @@
         "name" : "identifier",
         "type" : "token",
         "documentation" : "A patient identifier"
-      } ]
+      }, {
+        "name" : "startIndex",
+        "type" : "string"
+      }, {
+        "name" : "startIndex",
+        "type" : "string"} ]
```
karlmdavis: That... seems wrong. I wonder why HAPI thinks the param needs to be listed twice?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
jcgeer: Do you have any recommendation for where this might go? `TransformerUtils` might be a good place I thought, but I suppose it more aligns with the "utils" part of that, less so "transformer".

bluebutton-server-app/src/test/resources/endpoint-responses/metadata.json
jcgeer: Right, this one seems odd. I haven't looked at how the metadata is getting created, but I can tell it's from the fact that we're adding `startIndex` as an optional search parameter to both the search methods there. From that, I can surmise that it is pulling all the search parameters from each method, not bothering to see if they are unique or not.
I'll verify for sure if that is the case and see what we can do about it if you're not liking this look.

karlmdavis: I'd probably file an upstream bug on it. It's dumb, but likely won't cause any problems, I don't think.

@ekivemark You agree we don't need to solve this immediately? I can't imagine it'd break anything, but maybe I'm just not being imaginative enough today...

karlmdavis: @ekivemark (Context: we're adding paging to the `Patient` (and eventually also `Coverage`) backend resource types, and our auto-generated `/metadata` statement ends up with the `startIndex` parameter listed twice for some dumb reason.)

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
karlmdavis: Yeah, probably `TransformerUtils`. And maybe your now-moved `addPagingArgs(...)` method would be a better fit there, too. During the review, it looked a little odd to me, hanging out in `PagingArguments`.

I don't really have strong opinions on either, though.

jcgeer: Do you mean `addPagingLinks(...)`? I would agree, adding/creating the links don't really fit into `PagingArguments`, I just couldn't really think of a better place.

I'll through the `addResourcesToBundle(...)` to `TransformerUtils` for sure then, and I'll decide if I want to put the link stuff in there when I see how it looks, if it's no bother to you.

karlmdavis: > Do you mean `addPagingLinks(...)?`

Yeah, thanks.

> if it's no bother to you.

 👍

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PagingArguments.java
```diff
@@ -105,4 +108,64 @@ public int getStartIndex() {
...
+	 */
+	private String createPagingLink(String resource, String descriptor, String id, int startIndex, int theCount) {
+		StringBuilder b = new StringBuilder();
+		b.append(serverBase + resource);
+		b.append("_count=" + theCount);
+		b.append("&startIndex=" + startIndex);
+		b.append(descriptor + id);
+
+		return b.toString();
+	}
```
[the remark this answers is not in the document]
jcgeer: I've replaced most other strings with constants, but I cannot find anything for `startIndex`.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
jcgeer: I've moved a new version of `addResourcesToBundle(...)` into `TransformerUtils` as discussed.

After playing around with it, I've decided to keep the link creation/adding inside `PagingArguments` for now. Mostly because I'm using some of the instance variables/methods from `PagingArguments` in those two methods and it looks a bit cleaner keeping them there, and since they are at least related to paging.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
```diff
@@ -230,11 +228,11 @@ public Bundle findByPatient(
 			 * changes when this issue is resolved.
 			 */
 			int endIndex = Math.min(pagingArgs.getStartIndex() + pagingArgs.getPageSize(), eobs.size());
-			List<ExplanationOfBenefit> resources = eobs.subList(pagingArgs.getStartIndex(), endIndex);
-			bundle = addResourcesToBundle(bundle, resources);
-			pagingArgs.addPagingLinks(bundle, "/ExplanationOfBenefit?", "&patient=", beneficiaryId, eobs.size());
+			List<IBaseResource> resources = eobs.subList(pagingArgs.getStartIndex(), endIndex);
+			bundle = TransformerUtils.addResourcesToBundle(bundle, resources);
+			pagingArgs.addPagingLinks(bundle, "/ExplanationOfBenefit?", ExplanationOfBenefit.SP_PATIENT, beneficiaryId, eobs.size());
```
karlmdavis: If you passed `pagingArgs` into the `addPagingLinks` method, I think you could remove a decent chunk more code here: the if/else, the subList, etc.

bluebutton-server-app/src/test/resources/endpoint-responses/metadata.json
karlmdavis: @jcgeer Can you poke Mark on Slack/text/whatever to confirm that this isn't a problem?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExplanationOfBenefitResourceProvider.java
jcgeer: Hmm. Perhaps I'm not looking at this right. Adding `pagingArgs` to that method I can do. I'm not sure how I can get to a point to getting rid of the if/else and subList here, though. Those have more to do with deciding what resources to add to the bundle than they do adding/creating the links.

What I could do though is create a new method, say `createBundle` or something along those lines, pass in the `pagingArgs`, `resource` (EOB/Patient/Coverage), and the `identifier`, and have this code in there. Is that more what you're thinking?

bluebutton-server-app/src/test/resources/endpoint-responses/metadata.json
karlmdavis: @jcgeer and @ekivemark: Bump.

@jcgeer: If you don't hear back on this by tomorrow at noon, let's merge it anyways. I'll mark the PR as approved, pending the outcome of that.

ekivemark: It probably shouldn't cause a problem - unless someone submits a call with two different startIndex values in which case it would be interesting to see which one takes precedence.

I assume that we will get this resolved. 
I have no problem with adding paging to any resource since it is a standard feature of FHIR.

PR #141, opened 2019-04-11
BLUEBUTTON-802: Replace ignored fields with filler text

jcgeer:
First step, changes to the IT for a place to get a handle on the JSON
fields that will need to be updated. This in turn slightly changes the
format of all golden files. Committing them now so that in subsequent
commits we can look specifically at the fields that are being replaced.

### DIFF
--- a/bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
+++ b/bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
@@ -94,7 +94,7 @@ public final class EndpointJsonResponseComparatorIT {
 
 	private final String endpointId;
 	private final Supplier<String> endpointOperation;
-	private static final String ignoredFieldText = "IGNORED FIELD";
+	private static final String IGNORED_FIELD_TEXT = "IGNORED_FIELD";
 
 	/**
 	 * Parameterized test constructor: JUnit will construct a new instance of this
@@ -191,12 +191,12 @@ public final class EndpointJsonResponseComparatorIT {
 					if (fieldName == "url") {
 						// Only replace the port numbers on urls
 						String[] url = parent.get(fieldName).toString().split("[0-9]{4}");
-						String replacementUrl = url[0] + ignoredFieldText + url[1];
+						String replacementUrl = url[0] + IGNORED_FIELD_TEXT + url[1];
 						((ObjectNode) parent).put(fieldName, replacementUrl.substring(1, replacementUrl.length() - 1));
 					} else
-						((ObjectNode) parent).put(fieldName, ignoredFieldText);
+						((ObjectNode) parent).put(fieldName, IGNORED_FIELD_TEXT);
 			} else
-				((ObjectNode) parent).put(fieldName, ignoredFieldText);
+				((ObjectNode) parent).put(fieldName, IGNORED_FIELD_TEXT);
 		}
 
 		// Now, recursively invoke this method on all properties
--- a/bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
+++ b/bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
@@ -158,6 +158,9 @@ public final class EndpointJsonResponseComparatorIT {
 		replaceIgnoredFieldsWithFillerText(jsonNode, "url", Optional.of(Pattern.compile("https://localhost")));
 		replaceIgnoredFieldsWithFillerText(jsonNode, "lastUpdated", Optional.empty());
 
+		if (endpointId == "metadata")
+			replaceIgnoredFieldsWithFillerText(jsonNode, "date", Optional.empty());
+
 		String jsonResponse = null;
 		try {
 			jsonResponse = mapper.writerWithDefaultPrettyPrinter().writeValueAsString(jsonNode);
--- a/bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
+++ b/bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
@@ -14,6 +14,7 @@ import java.util.Collections;
 import java.util.Comparator;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
 import java.util.regex.Matcher;
@@ -144,7 +145,6 @@ public final class EndpointJsonResponseComparatorIT {
 		String endpointResponse = endpointOperation.get();
 
 		ObjectMapper mapper = new ObjectMapper();
-		mapper.writerWithDefaultPrettyPrinter();
 		JsonNode jsonNode = null;
 		try {
 			jsonNode = mapper.readTree(endpointResponse);
@@ -153,9 +153,10 @@ public final class EndpointJsonResponseComparatorIT {
 					"Unable to deserialize the following JSON content as tree: " + endpointResponse, e);
 		}
 
-		/*
-		 * TODO: Replace ignored fields with filler text here.
-		 */
+		replaceIgnoredFieldsWithFillerText(jsonNode, "id", Optional
+				.of(Pattern.compile("[A-Za-z0-9]{8}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{4}-[A-Za-z0-9]{12}")));
+		replaceIgnoredFieldsWithFillerText(jsonNode, "url", Optional.of(Pattern.compile("https://localhost")));
+		replaceIgnoredFieldsWithFillerText(jsonNode, "lastUpdated", Optional.empty());
 
 		String jsonResponse = null;
 		try {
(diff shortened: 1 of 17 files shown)


jcgeer:
@karlmdavis @dho937 @cthayes 

When reviewing this PR, it's probably easiest to look at the individual commits rather than all the changes at once. With the initial commit, I had to deserialize and then reserialize the JSON, and the methods I used to do so changed the formatting slightly to where all the response files have minor changes to the spacing. That will now be the spacing of the golden response files going forward.

In the two subsequent commits, https://github.com/CMSgov/bluebutton-data-server/pull/141/commits/f181ef9d4cdf78e05e7d16c56b9752ea1dd25944 and https://github.com/CMSgov/bluebutton-data-server/pull/141/commits/2fe34efe9a0b076b7b8effa329cd337761e03d81 you can see the changes I made to the IT to add filler text into those ignored fields, and the effect the changes had on the actual responses themselves in a much clearer way.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
```diff
@@ -187,9 +188,15 @@ private static void replaceIgnoredFieldsWithFillerText(JsonNode parent, String f
 				Pattern p = pattern.get();
 				Matcher m = p.matcher(parent.get(fieldName).toString());
 				if (m.find())
-					((ObjectNode) parent).put(fieldName, "IGNORED FIELD");
+					if (fieldName == "url") {
+						// Only replace the port numbers on urls
+						String[] url = parent.get(fieldName).toString().split("[0-9]{4}");
+						String replacementUrl = url[0] + ignoredFieldText + url[1];
```
karlmdavis: So, this isn't super robust or generally correct code. For example, given a URL such as, "`https://localhost:1234/foo/5678/bar`", it's going to behave unexpectedly.

**However**, I think all the bad edge cases would result in the IT failing, rather than false "everything still matches; carry on folks," results. So, doesn't need to be fixed.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
```diff
@@ -94,6 +94,7 @@
 
 	private final String endpointId;
 	private final Supplier<String> endpointOperation;
+	private static final String ignoredFieldText = "IGNORED FIELD";
```
karlmdavis: This is a constant, and should be capitalized, e.g. `IGNORED_FIELD_FILLER_TEXT`. And I'd probably also replace the space in the value with an underscore, because having spaces in the URL is fine for this use case, but still just looks weird.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
jcgeer: I see what you're saying, and it looks like my code was causing a tiny bit of an issue, chopping off the last bits of the URL as you can see in this commit https://github.com/CMSgov/bluebutton-data-server/pull/141/commits/391ef9e9832e71d4fde4159630856f014a6460ad. I've changed it a bit and I think that should solve any issues.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
```diff
@@ -189,14 +189,14 @@ private static void replaceIgnoredFieldsWithFillerText(JsonNode parent, String f
...
 				if (m.find())
 					if (fieldName == "url") {
-						// Only replace the port numbers on urls
-						String[] url = parent.get(fieldName).toString().split("[0-9]{4}");
-						String replacementUrl = url[0] + ignoredFieldText + url[1];
-						((ObjectNode) parent).put(fieldName, replacementUrl.substring(1, replacementUrl.length() - 1));
+						// Only replace the port numbers on url
+						String[] url = parent.get(fieldName).toString().split(pattern.get().toString());
+						String replacementUrl = "https://localhost:" + IGNORED_FIELD_TEXT + url[1];
+						((ObjectNode) parent).put(fieldName, replacementUrl.substring(0, replacementUrl.length() - 1));
```
karlmdavis: Can you look up regex capture groups and how you might use them here, instead? Worth learning about!

bluebutton-server-app/src/test/resources/endpoint-responses/eobByPatientIdAll.json
```diff
@@ -1,14 +1,14 @@
...
   "meta" : {
-    "lastUpdated" : "IGNORED FIELD"
+    "lastUpdated" : "IGNORED_FIELD"
   },
   "type" : "searchset",
   "total" : 8,
   "link" : [ {
     "relation" : "self",
-    "url" : "https://localhost:IGNORED FIELD/v1/fhir/ExplanationOfBenefit?_format=json&patient=Patient%2"
+    "url" : "https://localhost:IGNORED_FIELD/v1/fhir/ExplanationOfBenefit?_format=json&patient=Patient%2F567834"
```
karlmdavis: Oof. I hadn't noticed that before. Glad you fixed the parsing!

bluebutton-server-app/src/test/resources/endpoint-responses/patientSearchById.json
```diff
@@ -1,14 +1,14 @@
...
   "meta" : {
-    "lastUpdated" : "IGNORED FIELD"
+    "lastUpdated" : "IGNORED_FIELD"
   },
   "type" : "searchset",
   "total" : 1,
   "link" : [ {
     "relation" : "self",
-    "url" : "https://localhost:IGNORED FIELD/v1/fhir/Patient?_format=json&_id=%7"
+    "url" : "https://localhost:IGNORED_FIELD/v1/fhir/Patient?_format=json&_id=%7C567834"
```
karlmdavis: Any idea why the URLs have a "%7C" prefix? What's that resolve to?

jcgeer: I'm not actually sure what that is about. I've always seen a prefix to the ids in the requests, even running the separate ITs before I started work on this IT so I just assumed it was a part of how the URL was created. However, I really only noticed it in the EOBs, where you can see the prefix is "%2F".

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/utils/EndpointJsonResponseComparatorIT.java
jcgeer: Thanks for the tip! Definitely easier/cleaner looking than what I was doing and great to know for the future.

bluebutton-server-app/src/test/resources/endpoint-responses/patientSearchById.json
karlmdavis: It's a pipe character: `|`.

Which would make perfect sense if this was a FHIR _identifier_, where the format is "`system|id`". But IDs don't use that format. What is going on here...

jcgeer: Perhaps I'm not understanding this correctly. But if I look at the test for this (`patientSearchById`)...

The query's where clause is `where(Patient.RES_ID.exactly().systemAndIdentifier(null, beneficiary.getBeneficiaryId()))`. Looks like we are passing a system and id there, and being null there for system would cause just the pipe character, no?

I pulled that query straight from `PatientResourceProviderIT`. Are we doing that incorrectly then? Or am I not understanding quite right what you mean?

karlmdavis: So, we definitely have some bugs -- or at least oddities -- in our paging. See the `patient` query params here: >https://github.com/CMSgov/bluebutton-data-server/blob/master/bluebutton-server-app/src/test/resources/endpoint-responses/eobByPatientIdPaged.json#L9>. Some of them are formatted as `Patient/1234` and some are just `1234`.

And the pipe prefixes we're talking about here on `Patient` responses. I just went and looked through the `BeneficiaryTransformer` vs. the claim transformers, like `CarrierClaimTransformer`, and can't see any obvious reason that `Patient` `self` links get this weird prefix and EOBs don't. Unless it's something HAPI is doing because the `Patient.id` is all numeric, rather than alphanumeric?

Two action items, I think:

1. Make sure we have ITs that verify that **all** of the links resolve.
2. Experiment and/or dig through HAPI to see what causes that pipe prefix.

karlmdavis: Probably part of a separate ticket, btw.

karlmdavis: > The query's where clause is where(Patient.RES_ID.exactly().systemAndIdentifier(null, beneficiary.getBeneficiaryId())). Looks like we are passing a system and id there, and being null there for system would cause just the pipe character, no?

Interesting. So this isn't being generated by the HAPI response handling -- it's just using the URL that was called?

jcgeer: Probably... I'd have to dig deeper to figure out how exactly that is working. I know that I am for sure not creating any `self` links anywhere in the paging work I did, but didn't look into see exactly where it is coming from.

jcgeer: > Two action items, I think:
> 
> 1. Make sure we have ITs that verify that **all** of the links resolve.
> 2. Experiment and/or dig through HAPI to see what causes that pipe prefix.

I would agree. I'll go ahead and merge this then and can create some tickets to capture those two action items.

PR #142, opened 2019-04-26
BLUEBUTTON-786: Filtering other types

cthayes:
Add SAMHSA filtering and documentation for remaining claim types:
Outpatient
HHA
Hospice
SNF


dev/design-samhsa-filtering.md
```diff
@@ -34,49 +34,37 @@ The following table details all of the fields in our system which can contain SA
 |Claim Type(s)|Codeset|RIF Column(s)|RIF Column Format|Notes|FHIR EOB Field(s)|FHIR EOB Field Format|
 |---|---|---|---|---|---|---|
 |Carrier, DME|DRG|_(none)_|_(none)_|_(none)_|_(none)_|_(none)_|
-|Carrier, DME, Inpatient |CPT|`HCPCS_CD`|`\w{4-5}`|CPT codes are a subset ("Level I") of HCPCSs.|`ExplanationOfBenefit.item.service`|`"service": { "coding": [{ "system": "https://bluebutton.cms.gov/resources/codesystem/hcpcs", "version": "7", "code": "12345" }] }`|
-|Carrier, DME|CPT|`HCPCS_1ST_MDFR_CD`|`\w{2}`|Modifier codes provide additional information on the procedure performed.|_N/A_|_N/A_|
-|Carrier, DME|CPT|`HCPCS_2ND_MDFR_CD`|`\w{2}`|Modifier codes provide additional information on the procedure performed.|_N/A_|_N/A_|
+|Carrier, DME, Inpatient, Outpatient, SNF, HHA, Hospice|CPT|`HCPCS_CD`|`\w{4-5}`|CPT codes are a subset ("Level I") of HCPCSs.|`ExplanationOfBenefit.item.service`|`"service": { "coding": [{ "system": "https://bluebutton.cms.gov/resources/codesystem/hcpcs", "version": "7", "code": "12345" }] }`|
```
karlmdavis: Is there no `HCPCS_YR_CD` for the other claim types that have CPT codes? If that's the case, do you/Susie know how the HCPCS codeset year is supposed to be determined?

Note: This probably doesn't matter for SAMHSA filtering, but we should still investigate the discrepancy.

dev/design-samhsa-filtering.md
```diff
@@ -34,49 +34,37 @@ The following table details all of the fields in our system which can contain SA
...
+|Inpatient, SNF|ICD Diagnosis|`ADMTG_DGNS_CD`|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD` but without `type`)_|
+|Inpatient, SNF|ICD Diagnosis|`ADMTG_DGNS_VRSN_CD`|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|
+|Inpatient, Outpatient, SNF, HHA, Hospice|ICD Diagnosis|`ICD_DGNS_CD[1-25]`|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD` but without `type`)_|
+|Inpatient, Outpatient, SNF, HHA, Hospice|ICD Diagnosis|`ICD_DGNS_VRSN_CD[1-25]`|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|
+|Inpatient, Outpatient, SNF, HHA, Hospice|ICD Diagnosis|`FST_DGNS_E_CD`|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD` but without `type`)_|
```
karlmdavis: We don't have to filter on these, right? I'm _pretty_ sure the list of ICD codes we've got don't include any E codes, but did you double-check?

dev/design-samhsa-filtering.md
```diff
@@ -34,49 +34,37 @@ The following table details all of the fields in our system which can contain SA
...
+|Inpatient, Outpatient, SNF|ICD Procedure|`ICD_PRCDR_VRSN_CD[1-25]`|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|
+|Outpatient|ICD Diagnosis|`RSN_VISIT_CD[1-3]`|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD`)_|_(same as `PRNCPAL_DGNS_CD` but without `type`)_|
+|Outpatient|ICD Diagnosis|`RSN_VISIT_VRSN_CD[1-3]`|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|_(same as `PRNCPAL_DGNS_VRSN_CD`)_|
+|Part D Events|DRG|_(none)_|_(none)_|_(none)_|_(none)_|_(none)_|
+|Part D Events|CPT|_(none)_|_(none)_|_(none)_|_(none)_|_(none)_|
+|Part D Events|ICD Diagnosis|_(none)_|_(none)_|_(none)_|_(none)_|_(none)_|
+|Part D Events|ICD Procedure|_(none)_|_(none)_|_(none)_|_(none)_|_(none)_|
```
karlmdavis: In general, this is great! But I think -- now that it's done -- it's hard to tell exactly which of these possible fields we're _actually_ filtering on. Can you add a column that indicates that, and have Susie review it?

Or am I being silly, and just not seeing how to determine that from the info in the table?

dev/design-samhsa-filtering.md
cthayes: Yeah there's nothing like CARR_CLM_HCPCS_YR_CD for the other claim types. The transformers pass an empty optional as the year so it's just not set HCPCS coding.

dev/design-samhsa-filtering.md
cthayes: Yeah Susie said there is nothing in the E codes to check, but since the E codes are grouped with the other diagnosis codes they will be checked by default. So it's useful to know that if there is anything in the E codes in the future all we need to update is the CSV file.

dev/design-samhsa-filtering.md
cthayes: Is this comment in regards to PDE specifically? If so, we're not filtering PDE claims at all.

karlmdavis: Whoops, that was definitely unclear. Apologies: I meant the whole table.

cthayes: Besides the rows listed with none the only field we're not filtering is the diagnosis e codes right? I added something in the notes column. Are you thinking something else is needed here?

karlmdavis:
Note that Jenkins is unhappy, so you might have to go kick it.

PR #143, opened 2019-04-29
BLUEBUTTON-865: Adding plaintext HICN and MBI to patient resource

jcgeer:
Adding two additional identifiers to the patient resource when querying
for BCDA users (includeIdentifiers=true).

BLUEBUTTON-865

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
@@ -25,13 +25,14 @@ final class BeneficiaryTransformer {
 	 *            the {@link MetricRegistry} to use
 	 * @param beneficiary
 	 *            the CCW {@link Beneficiary} to transform
+	 * @param includeIdentifiers
 	 * @return a FHIR {@link Patient} resource that represents the specified
 	 *         {@link Beneficiary}
 	 */
-	public static Patient transform(MetricRegistry metricRegistry, Beneficiary beneficiary) {
+	public static Patient transform(MetricRegistry metricRegistry, Beneficiary beneficiary, String includeIdentifiers) {
 		Timer.Context timer = metricRegistry
 				.timer(MetricRegistry.name(BeneficiaryTransformer.class.getSimpleName(), "transform")).time();
-		Patient patient = transform(beneficiary);
+		Patient patient = transform(beneficiary, includeIdentifiers);
 		timer.stop();
 
 		return patient;
@@ -40,10 +41,11 @@ final class BeneficiaryTransformer {
 	/**
 	 * @param beneficiary
 	 *            the CCW {@link Beneficiary} to transform
+	 * @param includeIdentifiers
 	 * @return a FHIR {@link Patient} resource that represents the specified
 	 *         {@link Beneficiary}
 	 */
-	private static Patient transform(Beneficiary beneficiary) {
+	private static Patient transform(Beneficiary beneficiary, String includeIdentifiers) {
 		Objects.requireNonNull(beneficiary);
 
 		Patient patient = new Patient();
@@ -54,15 +56,20 @@ final class BeneficiaryTransformer {
 		patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_BENE_HICN_HASH)
 				.setValue(beneficiary.getHicn());
 
-		patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_BENE_HICN_UNHASHED)
-				.setValue(beneficiary.getHicnUnhashed().get());
-		for (BeneficiaryHistory beneHistory : beneficiary.getBeneficiaryHistories()) {
+		if (Boolean.parseBoolean(includeIdentifiers) == true) {
 			patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_BENE_HICN_UNHASHED)
-					.setValue(beneHistory.getHicnUnhashed().get());
-		}
-		for (MedicareBeneficiaryIdHistory mbiHistory : beneficiary.getMedicareBeneficiaryIdHistories()) {
-			patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_MEDICARE_BENEFICIARY_ID)
-					.setValue(mbiHistory.getMedicareBeneficiaryId().get());
+					.setValue(beneficiary.getHicnUnhashed().get());
+			for (BeneficiaryHistory beneHistory : beneficiary.getBeneficiaryHistories()) {
+				patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_BENE_HICN_UNHASHED)
+						.setValue(beneHistory.getHicnUnhashed().get());
+			}
+
+			patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_MEDICARE_BENEFICIARY_ID_UNHASHED)
+					.setValue(beneficiary.getMedicareBeneficiaryId().get());
+			for (MedicareBeneficiaryIdHistory mbiHistory : beneficiary.getMedicareBeneficiaryIdHistories()) {
+				patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_MEDICARE_BENEFICIARY_ID_UNHASHED)
+						.setValue(mbiHistory.getMedicareBeneficiaryId().get());
+			}
 		}
 
 		patient.addAddress().setState(beneficiary.getStateCode()).setDistrict(beneficiary.getCountyCode())
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
@@ -17,6 +17,7 @@ import gov.hhs.cms.bluebutton.data.model.rif.Beneficiary;
 import gov.hhs.cms.bluebutton.data.model.rif.BeneficiaryHistory;
 import gov.hhs.cms.bluebutton.data.model.rif.MedicareBeneficiaryIdHistory;
 import gov.hhs.cms.bluebutton.data.model.rif.parse.InvalidRifValueException;
+import gov.hhs.cms.bluebutton.server.app.stu3.providers.PatientResourceProvider.IncludeIdentifiersMode;
 
 /**
  * Transforms CCW {@link Beneficiary} instances into FHIR {@link Patient}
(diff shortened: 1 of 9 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -305,4 +313,24 @@ private Patient queryDatabaseByHicnHash(String hicnHash) {
 		Patient patient = BeneficiaryTransformer.transform(metricRegistry, beneficiary);
 		return patient;
 	}
+
+	/**
+	 * @param hicnHash
+	 *            the {@link Beneficiary#getHicn()} hash value to match
+	 */
+	private void addUnhashedHicnAndMbi(List<IBaseResource> patients) {
```
karlmdavis: What are your thoughts on this: can and should we try to avoid issuing a separate DB query here, by doing a joined query for the initial lookup?

nickrobison-usds:
Following up on @karlmdavis initial comment. It would be great to have historical IDs, along with the period for which those ids were valid. Do we also have a way to indicate which IDs are currently in force?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
```diff
@@ -23,13 +25,14 @@
 	 *            the {@link MetricRegistry} to use
 	 * @param beneficiary
 	 *            the CCW {@link Beneficiary} to transform
+	 * @param includeIdentifiers
 	 * @return a FHIR {@link Patient} resource that represents the specified
 	 *         {@link Beneficiary}
 	 */
-	public static Patient transform(MetricRegistry metricRegistry, Beneficiary beneficiary) {
+	public static Patient transform(MetricRegistry metricRegistry, Beneficiary beneficiary, String includeIdentifiers) {
```
karlmdavis: Can you make `includeIdentifiers` a `boolean`, at least here?

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerConstants.java
```diff
@@ -223,6 +223,18 @@
 	 * now.
 	 */
 	public static final String CODING_BBAPI_BENE_HICN_HASH = BASE_URL_BBAPI_RESOURCES + "/identifier/hicn-hash";
+	
+	/**
+	 * The {@link Identifier#getSystem()} used in {@link Patient} resources to
+	 * store the unhashed version of each Medicare beneficiaries' HICN.
+	 */
+	public static final String CODING_BBAPI_BENE_HICN_UNHASHED = BASE_URL_BBAPI_RESOURCES + "/identifier/hicn-unhashed";
```
karlmdavis: Per <https://www.hl7.org/fhir/identifier-registry.html>, this should probably be "`http://hl7.org/fhir/sid/us-medicare`".

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerConstants.java
```diff
@@ -223,6 +223,18 @@
...
+
+	/**
+	 * The {@link Identifier#getSystem()} used in {@link Patient} resources to
+	 * store the each Medicare beneficiaries' medicare beneficiary id.
+	 */
+	public static final String CODING_BBAPI_MEDICARE_BENEFICIARY_ID_UNHASHED = BASE_URL_BBAPI_RESOURCES + "/identifier/mbi-unhashed";
```
karlmdavis: Per <https://www.hl7.org/fhir/identifier-registry.html>, this should probably be "`http://hl7.org/fhir/sid/us-mbi`".

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
```diff
@@ -52,6 +56,22 @@ private static Patient transform(Beneficiary beneficiary) {
 		patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_BENE_HICN_HASH)
 				.setValue(beneficiary.getHicn());
 
+		if (Boolean.parseBoolean(includeIdentifiers) == true) {
+			patient.addIdentifier().setSystem(TransformerConstants.CODING_BBAPI_BENE_HICN_UNHASHED)
+					.setValue(beneficiary.getHicnUnhashed().get());
+			for (BeneficiaryHistory beneHistory : beneficiary.getBeneficiaryHistories()) {
```
karlmdavis: Do we know that `BeneficiaryHistory` will always contain the most current HICN, as well? Or do we need to grab that from `Beneficiary.hicnUnhashed` and add it, too?

Also: we should probably do a couple other things to make this more useful for end users:
* Verify that the list won't include any duplicates.
* Mark each entry with a timestamp, or maybe better to just mark the most current HICN as such?

(This all applies to MBIs, as well.)

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -266,6 +272,8 @@ private Patient queryDatabaseByHicnHash(String hicnHash) {
 				.timer(MetricRegistry.name(getClass().getSimpleName(), "query", "bene_by_hicn", "current")).time();
 		CriteriaQuery<String> beneHicnQuery = builder.createQuery(String.class);
 		Root<Beneficiary> beneHicnQueryRoot = beneHicnQuery.from(Beneficiary.class);
+		beneHicnQueryRoot.join(Beneficiary_.beneficiaryHistories);
+		beneHicnQueryRoot.join(Beneficiary_.medicareBeneficiaryIdHistories);
```
karlmdavis: I love that these are joined in a single query -- great work!

Do you think there's a way to update the `.where(...)` to also include the HICNs from `BeneficiaryHistory`, as well -- in the same query? (Fine to leave that for a future optimization.)

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/ExtraParamsInterceptor.java
```diff
@@ -0,0 +1,28 @@
+package gov.hhs.cms.bluebutton.server.app.stu3.providers;
+
+import java.io.IOException;
+
+import ca.uhn.fhir.rest.client.api.IClientInterceptor;
+import ca.uhn.fhir.rest.client.api.IHttpRequest;
+import ca.uhn.fhir.rest.client.api.IHttpResponse;
+
+public class ExtraParamsInterceptor implements IClientInterceptor {
```
karlmdavis: I'd add a comment explaining why this silliness is needed.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProviderIT.java
```diff
@@ -40,6 +42,50 @@ public void readExistingPatient() {
...
+		ExtraParamsInterceptor extraParamsInterceptor = new ExtraParamsInterceptor();
+		extraParamsInterceptor.setIncludeIdentifiers("false");
+		fhirClient.registerInterceptor(extraParamsInterceptor);
+
+		Beneficiary beneficiary = loadedRecords.stream().filter(r -> r instanceof Beneficiary).map(r -> (Beneficiary) r)
+				.findFirst().get();
+		Patient patient = fhirClient.read().resource(Patient.class).withId(beneficiary.getBeneficiaryId()).execute();
+
+		Assert.assertNotNull(patient);
+		BeneficiaryTransformerTest.assertMatches(beneficiary, patient);
```
karlmdavis: I think you also need an assertion that the identifiers aren't present?

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProviderIT.java
```diff
@@ -40,6 +42,50 @@ public void readExistingPatient() {
...
+		ExtraParamsInterceptor extraParamsInterceptor = new ExtraParamsInterceptor();
+		extraParamsInterceptor.setIncludeIdentifiers("true");
+		fhirClient.registerInterceptor(extraParamsInterceptor);
+
+		Beneficiary beneficiary = loadedRecords.stream().filter(r -> r instanceof Beneficiary).map(r -> (Beneficiary) r)
+				.findFirst().get();
+		Patient patient = fhirClient.read().resource(Patient.class).withId(beneficiary.getBeneficiaryId()).execute();
+
+		Assert.assertNotNull(patient);
+		BeneficiaryTransformerTest.assertMatches(beneficiary, patient);
```
karlmdavis: I think you also need an assertion that the identifiers are present?

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProviderIT.java
```diff
@@ -166,6 +254,49 @@ public void searchForExistingPatientByHicnHash() {
...
 	}
 
+	/**
+	 * Verifies that
+	 * {@link PatientResourceProvider#searchByIdentifier(ca.uhn.fhir.rest.param.TokenParam)}
+	 * works as expected for a {@link Patient} that does exist in the DB, including
+	 * identifiers to return the unhashed HICN and MBI.
+	 */
+	@Test
+	public void searchForExistingPatientByHicnHashIncludeIdentifiers() {
```
karlmdavis: I know it's silly, but I'd also add `searchForExistingPatientByHicnHashIncludeIdentifiersFalse`, as them not showing up is a major security concern.

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProviderIT.java
```diff
@@ -85,6 +131,48 @@ public void searchForExistingPatientByLogicalId() {
...
 	}
 
+	/**
+	 * Verifies that
+	 * {@link PatientResourceProvider#searchByLogicalId(ca.uhn.fhir.rest.param.TokenParam)}
+	 * works as expected for a {@link Patient} that does exist in the DB, including
+	 * identifiers to return the unhashed HICN and MBI.
+	 */
+	@Test
+	public void searchForExistingPatientByLogicalIdIncludeIdentifiers() {
```
karlmdavis: I know it's silly, but I'd also add `searchForExistingPatientByLogicalIdIncludeIdentifiersFalse`, as them not showing up is a major security concern.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
jcgeer: At least in our test data, it appears that `BeneficiaryHistory` and `MedicareBeneficiaryIdHistory` did not contain the most current HICN/MBIs, and pulling from `Beneficiary.hicnUnhashed` seemed necessary. Where could I go/who to ask to check if that is the case always?

I've added some additional code to verify we won't be including duplicates. As far as the timestamp/marking most current HICN/MBI... I could add a period to the identifier and that should work for MBI at least since `MedicareBeneficiaryIdHistory` has an effective and end date. For `BeneficiaryHistory`, however, it doesn't look like we're capturing that data. Do you have any suggestions of how to mark the *current* one as such?

jcgeer: A quick look at the data, it doesn't seem like the date fields are always present in `MedicareBeneficiaryIdHistory`, so I'm not sure we can use that effectively. Did you have a different kind of timestamp in mind?

jcgeer:
@karlmdavis Ready for another review.

PR #144, opened 2019-04-30
Updating to check for correct sex value characters

jcgeer:
Changing to check the sex field for the correct values of 0/1/2 (as the
data is stored in our database) versus M/F.

BLUEBUTTON-898

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
@@ -9,9 +9,9 @@ import org.hl7.fhir.dstu3.model.Patient;
 import com.codahale.metrics.MetricRegistry;
 import com.codahale.metrics.Timer;
 
-import ca.uhn.fhir.rest.server.exceptions.InvalidRequestException;
 import gov.hhs.cms.bluebutton.data.codebook.data.CcwCodebookVariable;
 import gov.hhs.cms.bluebutton.data.model.rif.Beneficiary;
+import gov.hhs.cms.bluebutton.data.model.rif.parse.InvalidRifValueException;
 
 /**
  * Transforms CCW {@link Beneficiary} instances into FHIR {@link Patient}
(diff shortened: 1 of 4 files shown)


jcgeer:
> Why doesn't this change require the approved response JSON files for `Patient` to be updated?

The `BeneficiaryTransformer` was looking for 'M/F' originally, we see that, and it wasn't working for actual data out in CCW because that is being stored as '0/1/2'. However, our sample files were using 'M/F', so the check was passing correctly and the response files were made with the appropriate gender descriptions.

karlmdavis:
Ahh, I see now. That's interesting.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformer.java
```diff
@@ -58,17 +59,16 @@ private static Patient transform(Beneficiary beneficiary) {
...
+		else if (sex == Sex.FEMALE.getCode())
 			patient.setGender((AdministrativeGender.FEMALE));
-			break;
-		default:
+		else if (sex == Sex.UNKNOWN.getCode())
 			patient.setGender((AdministrativeGender.UNKNOWN));
-			break;
-		}
+		else
+			throw new InvalidRequestException(
```
karlmdavis: Deffo not the right exception: this one, generally, would indicate that the person making the request did a goof.

What we want here, instead, is something that indicates that our understanding of our DB is incorrect. Not sure what that'd be off the top of my head, but if you look through the other transformers, I imagine you'll find some examples. If you're _really_ lucky, those examples will be consistent with each other and not just throwing a generic `RuntimeException`.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/Sex.java
```diff
@@ -0,0 +1,15 @@
+package gov.hhs.cms.bluebutton.server.app.stu3.providers;
+
+public enum Sex {
```
karlmdavis: Could you slap some JavaDocs on this bad boy?

bluebutton-server-app/src/test/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/BeneficiaryTransformerTest.java
```diff
@@ -74,9 +74,9 @@ static void assertMatches(Beneficiary beneficiary, Patient patient) {
 		Assert.assertEquals(beneficiary.getCountyCode(), patient.getAddress().get(0).getDistrict());
 		Assert.assertEquals(beneficiary.getPostalCode(), patient.getAddress().get(0).getPostalCode());
 		Assert.assertEquals(Date.valueOf(beneficiary.getBirthDate()), patient.getBirthDate());
-		if (beneficiary.getSex() == '1')
+		if (beneficiary.getSex() == Sex.MALE.getCode())
 			Assert.assertEquals("MALE", patient.getGender().toString().trim());
```
karlmdavis: Is there a HAPI/FHIR enum you can use here for the test comparisons?

PR #152, opened 2019-07-02
BLUEBUTTON-1033: Fixed a bug in the FDA drug lookup

karlmdavis:
It was failing for bad data, which means it failed for just about every record in our TEST environment.

https://jira.cms.gov/browse/BLUEBUTTON-1033

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
@@ -2854,7 +2854,11 @@ public final class TransformerUtils {
 	 */
 	public static String retrieveFDADrugCodeDisplay(String claimDrugCode) {
 
-		if (claimDrugCode.isEmpty())
+		/*
+		 * Handle bad data (e.g. our random test data) if drug code is empty or
+		 * length is less than 9 characters
+		 */
+		if (claimDrugCode.isEmpty() || claimDrugCode.length() < 9)
 			return null;
 
 		/*
@@ -2869,10 +2873,8 @@ public final class TransformerUtils {
 		}
 
 		String claimDrugCodeReformatted = null;
-		// Handle bad data (e.g. our random test data).
-		if (claimDrugCode.length() == 9) {
-			claimDrugCodeReformatted = claimDrugCode.substring(0, 5) + "-" + claimDrugCode.substring(5, 9);
-		}
+
+		claimDrugCodeReformatted = claimDrugCode.substring(0, 5) + "-" + claimDrugCode.substring(5, 9);
 
 		if (ndcProductMap.containsKey(claimDrugCodeReformatted)) {
 			String ndcSubstanceName = ndcProductMap.get(claimDrugCodeReformatted);
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
@@ -2868,18 +2868,22 @@ public final class TransformerUtils {
 			ndcProductMap = readFDADrugCodeFile();
 		}
 
-		String claimDrugCodeReformatted = claimDrugCode.substring(0, 5) + "-" + claimDrugCode.substring(5, 9);
-	    
+		String claimDrugCodeReformatted = null;
+		// Handle bad data (e.g. our random test data).
+		if (claimDrugCode.length() == 9) {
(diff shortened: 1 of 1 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2868,18 +2868,22 @@ public static String retrieveFDADrugCodeDisplay(String claimDrugCode) {
 			ndcProductMap = readFDADrugCodeFile();
 		}
 
-		String claimDrugCodeReformatted = claimDrugCode.substring(0, 5) + "-" + claimDrugCode.substring(5, 9);
-	    
+		String claimDrugCodeReformatted = null;
+		// Handle bad data (e.g. our random test data).
+		if (claimDrugCode.length() == 9) {
```
karlmdavis: Please check my math here: I'm pretty sick today and very well may have screwed up counting.

bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/TransformerUtils.java
```diff
@@ -2868,18 +2868,22 @@ public static String retrieveFDADrugCodeDisplay(String claimDrugCode) {
 			ndcProductMap = readFDADrugCodeFile();
 		}
 
-		String claimDrugCodeReformatted = claimDrugCode.substring(0, 5) + "-" + claimDrugCode.substring(5, 9);
-	    
+		String claimDrugCodeReformatted = null;
+		// Handle bad data (e.g. our random test data).
```
karlmdavis: @dho937 You're the expert here... can you help me understand what the right thing to do here is?

* Our random test data has NDC codes with less than 9 characters, which blows up the old code here (trying to get a substring past the end of the string).
* I figured, "okay, let's just make sure that the input is 9 characters long before reformatting it." That should fix it.
* Now we're getting test failures from `PartDEventTransformerTest.assertMatches:161` because the `SAMPLE_A` data has an NDC code with _11_ characters: a lookup that worked before is now failing.
    * Should `51270012299` get transformed & truncated to `51270-0122` so that it resolves to "`HASOL Anagen Hair Tonic - DEXPANTHENOL; MENTHOL; SALICYLIC ACID`"? Or was that a bug, previously?

dho937: @karlmdavis   I reviewed this change incorrectly before.  Not sure what was failing before but the only claim type that we are running a JUnit for to test how the NDC code is parsed is for the Part D claim type which has an 11 digit NDC code.  Yes, 51270012299 should be formatted to 51270-0122 which resolves to "HASOL Anagen Hair Tonic - DEXPANTHENOL; MENTHOL; SALICYLIC ACID".     I don't think we need the length check that you have added to the code.  I removed the length code and it builds fine.    Yes, some of our other sample data may have NDC lengths < 11 but we should make those codes more realistic if we are going to run them through our JUnit tests.

dho937: @karlmdavis   Were the new MCT PDE records that you added for MCT the ones that were failing?   I imagine they were as I think about this more....

karlmdavis: > I don't think we need the length check that you have added to the code. I removed the length code and it builds fine. Yes, some of our other sample data may have NDC lengths < 11 but we should make those codes more realistic if we are going to run them through our JUnit tests.

The problem isn't with codes that are longer than characters. It's that calling `claimDrugCode.substring(5, 9)` throws an `IndexOutOfBoundsException` for strings _shorter_ than 9 characters, like much of our garbage random data in TEST. That's where I'm seeing the failures.

> Were the new MCT PDE records that you added for MCT the ones that were failing?

Nope, those aren't loaded in TEST, which is where I was seeing this.

> 51270012299 should be formatted to 51270-0122

That's interesting! What are the last two characters on it that we're ignoring for, then?

dho937: @karlmdavis   In Test...all the PDE NDC's are 11 long; all the DME NDC's are 9 long and there are 4 claims for the same bene id (10000047) of Carrier claims with the NDC being 1 long.   I have updated the TransformerUtils class to handle the bad length NDC's and pushed this code up to the branch in GitHub.

karlmdavis: Wow!

1. Great work researching and fixing that!
2. I find it fascinating that someone managed to find the one `BENE_ID` where the data's borked, such that we had those errors in our logs.

I love messy stuff like this. Thanks for the assist!

PR #153, opened 2019-07-17
BLUEBUTTON-1061: Fixing NoSuchElementException PROD error

jcgeer:
Fixing our code to account for missing data for diagnosis admitting code
in the SNFClaimTransformer.


jcgeer:
@karlmdavis I checked the `InpatientClaimTransformer` as well as `TransformerUtils.extractDiagnoses1Thru12` where we are adding diagnoses and they all had a handy `Consumer` class setup to do the same check repeatedly.

Since we're only doing it once here I didn't think it necessary to do it the same way and instead went with your initial suggestion.

karlmdavis:
Can you get this merged and deployed ASAP?

PR #154, opened 2019-07-22
Bluebutton-983: new code to handle hicn pointing to more than one bene

dho937:
@karlmdavis   Can I get a review on this ASAP?  This is part of the code I need to move to production soon to fix the 500 errors we are getting where a HICN points to more than one bene id.  The JUnit I created is a little complex but don't have time to refactor now as this needs to get to production soon.  Thanks.

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
@@ -1,15 +1,16 @@
 package gov.hhs.cms.bluebutton.server.app.stu3.providers;
 
+import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
 import javax.persistence.NoResultException;
-import javax.persistence.NonUniqueResultException;
 import javax.persistence.PersistenceContext;
 import javax.persistence.criteria.CriteriaBuilder;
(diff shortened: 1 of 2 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -305,4 +346,50 @@ private Patient queryDatabaseByHicnHash(String hicnHash) {
 		Patient patient = BeneficiaryTransformer.transform(metricRegistry, beneficiary);
 		return patient;
 	}
+
+	/**
+	 * 
+	 * Following method will bring back the bene id that has the most recent
+	 * rfrnc_yr since the hicn points to more than one bene id in the
+	 * Beneficiaries table
```
karlmdavis: How do we know that this the right thing to do in this circumstance?

Also: are there any cases where a single HICN points to _multiple_ bene records with the same `RFRNC_YR`? If so this code won't notice; it'll just retrurn the first of those that it happens to come across.

karlmdavis (commented):
@dho937 Great work!

I noted a couple of important questions in a comment on your `getMaxReferenceYearForDupBenes(...)` method. We'll need to get those answers if we don't have them, and document them, e.g. in a comment.

Also: given the changes I made yesterday in https://github.com/CMSgov/bluebutton-data-server/pull/157, this will need a fair bit of rework, but should also end up much simpler (as there's a spot in the code where we already have all of the matching bene records and their data available, so you won't need to run any queries).

jcgeer (approved):
Looks good to me, but I don't see anywhere you've answered Karl's questions concerning always returning the bene with the latest `RFRNC_YR`.

karlmdavis:
Closing this PR, as the project has now been migrated into a new monorepo: https://github.com/CMSgov/beneficiary-fhir-data

PR #157, opened 2019-07-26
BLUEBUTTON-865: return plaintext HICNs and MBIs for BCDA (Round 2)

karlmdavis:
This is Round 2 for the changes from this earlier (reverted) PR: https://github.com/CMSgov/bluebutton-data-server/pull/143. Basically, we're adding an option to include plaintext HICNs and MBIs in the `Patient.identifier` results.

Originally, the PR was reverted as it needed a DB index that wasn't present. That index was added in https://github.com/CMSgov/bluebutton-data-model/pull/77, and so this enhancement should now be safe to deploy.

This PR is a revert-of-the-revert, with some additional tweaks:

* Improve performance a bit by going from 3 DB queries down to 1.
* Try harder to not send plaintext HICNs and MBIs to the transformer, when they're not needed.
* Tweak the API changelog a bit.
* Added `EndpointJsonResponseComparatorIT` test cases to cover the new query types.

https://jira.cms.gov/browse/BLUEBUTTON-865

See also: https://jira.cms.gov/browse/BLUEBUTTON-1110

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
@@ -103,7 +103,7 @@ public final class PatientResourceProvider implements IResourceProvider {
 	 *         <code>null</code> if none exists.
 	 */
 	@Read(version = false)
-	public Patient read(@IdParam IdType patientId) {
+	public Patient read(@IdParam IdType patientId, RequestDetails requestDetails) {
 		if (patientId == null)
 			throw new IllegalArgumentException();
 		if (patientId.getVersionIdPartAsLong() != null)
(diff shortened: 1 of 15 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -256,53 +273,92 @@ private Patient queryDatabaseByHicnHash(String hicnHash) {
...
+		 *
+		 * SELECT
+		 *     *   -- Retrieved columns are dynamic, based on JPA fetch groups and IncludeIdentifiers.
+		 *   FROM "Beneficiaries"
+		 *   INNER JOIN "BeneficiariesHistory" ON "Beneficiaries"."beneficiaryId" = "BeneficiariesHistory"."beneficiaryId"
+		 *   INNER JOIN "MedicareBeneficiaryIdHistory" ON "Beneficiaries"."beneficiaryId" = "MedicareBeneficiaryIdHistory"."beneficiaryId"  -- Might be omitted, if JPA is smart enough.
+		 *   WHERE
+		 *     "Beneficiaries"."hicn" = $1
+		 *     OR "BeneficiariesHistory"."hicn" = $1
+		 *
```
dho937: I think these INNER JOIN's need to be LEFT JOIN's.   I ran a couple SQL queries where the hicn was in the beneficiaries table but NOT in the BeneficiariesHistory table and I got no results by using the query above.     Still trying to test the scenario of a hicn only being in the BeneficiariesHistory table and not in the Beneficiaries table....   Not an expert on how the joins work in JPA....

karlmdavis: The JavaDoc for `From.join(SetAttribute)` says this:

> Create an inner join to the specified Set-valued attribute.

So great catch!

Let me go see why our ITs didn't catch this, because that's a bad miss...

karlmdavis: The join is `ON Beneficiaries.beneficiaryId = BeneficiariesHistory.beneficiaryId`, so I think the fail case is where we have a `Beneficiary` that doesn't have any history records at all.

`SAMPLE_A` doesn't cover that case, so I think that's the problem. Let me see if I can prove it...

dho937: Yes...that is one of the scenarios I have created with my additional test data....SAMPLE_A is basically happy path which doesn't help here...

karlmdavis: Confirmed: just added a `PatientResourceProviderIT.searchForExistingPatientWithNoHistory()` test case that fails.

Bravo! 👏 

I'll push the test and the fix in a moment.

karlmdavis: Fix pushed. Going to re-request review.

Many, many kudos for catching this bug during review: you absolutely prevented a nasty incident that would have gone right into PROD!

karlmdavis:
@dho937 Thanks for checking. I believe that `PatientResourceProviderIT.searchForExistingPatientByHistoricalHicnHash()` covers the other edge case you mentioned.

PR #158, opened 2019-08-05
BLUEBUTTON-865: return plaintext HICNs and MBIs for BCDA (round 3)

karlmdavis:
This is Round 3 for the changes from this earlier (reverted) PR: https://github.com/CMSgov/bluebutton-data-server/pull/143. Basically, we're adding an option to include plaintext HICNs and MBIs in the `Patient.identifier` results.

Originally, the PR was reverted as it needed a DB index that wasn't present. That index was added in https://github.com/CMSgov/bluebutton-data-model/pull/77, and so this enhancement should now be safe to deploy.

The "round 2" attempt from this PR also failed in PROD: https://github.com/CMSgov/bluebutton-data-server/pull/157. It turns out that in order to properly respond to the query used there, the DB had to first `LEFT JOIN` all rows in `BeneficiariesHistory` to all rows in `Beneficiaries`, which takes forever. We avoid that problem here by manually doing that join/lookup via a preliminary query.

This PR is a revert-of-the-revert, with some additional tweaks:

* Improve performance a bit by going from 3 DB queries down to 2.
* Try harder to not send plaintext HICNs and MBIs to the transformer, when they're not needed.
* Tweak the API changelog a bit.
* Added `EndpointJsonResponseComparatorIT` test cases to cover the new query types.

https://jira.cms.gov/browse/BLUEBUTTON-865

See also: https://jira.cms.gov/browse/BLUEBUTTON-1110

### DIFF
--- a/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
+++ b/bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
@@ -1,9 +1,11 @@
 package gov.hhs.cms.bluebutton.server.app.stu3.providers;
 
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Optional;
+import java.util.Set;
 
 import javax.inject.Inject;
 import javax.persistence.EntityManager;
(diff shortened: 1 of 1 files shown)


bluebutton-server-app/src/main/java/gov/hhs/cms/bluebutton/server/app/stu3/providers/PatientResourceProvider.java
```diff
@@ -272,63 +270,100 @@ private Patient queryDatabaseByHicnHash(String hicnHash, RequestDetails requestD
...
+			beneMatches.where(builder.or(beneHicnMatches, beneHistoryHicnMatches));
+		} else {
+			beneMatches.where(beneHicnMatches);
 		}
-		criteria.select(rootBenes);
-		Predicate beneHicnMatches = builder.equal(rootBenes.get(Beneficiary_.hicn), hicnHash);
-		Predicate beneHistoryHicnMatches = builder.equal(joinHistory.get(BeneficiaryHistory_.hicn), hicnHash);
-		criteria.where(builder.or(beneHicnMatches, beneHistoryHicnMatches));
-		List<Beneficiary> beneficiaries;
+		List<Beneficiary> macthingBenes;
```
dho937: misspelled "matching"...other than that..looks fine...

karlmdavis:
Going to merge, even though the last changes haven't been reviewed -- they're nonfunctional.

Also, going to have to test this by hand before deploying, anyways.

PR #166, opened 2019-08-15
Use different JPA config setting

cthayes:
BLUEBUTTON-1189


cthayes:
@karlmdavis Looks like we are now seeing the following 500 error. Waiting on confirmation on the DB side.
HTTP 500 Internal Server Error: Failed to call access method: javax.persistence.PersistenceException: org.hibernate.exception.GenericJDBCException: could not extract ResultSet

cthayes:
@karlmdavis Brandon verified that no requests ran over the 2 second timeout in test.

